PR #14, opened 2021-10-27
Add test data, remove fasta and igenomes params

erikrikarddaniel:
I added the same test data that I'm using for #magmap.
I also removed the `fasta` and `igenomes` params as I see no use for them now. (`fasta` could be the stable RNA, but I prefer using a more explicit name for that. (In fact, I removed the whole reference genome section from `nextflow_schema.json`.)


emnilsson (approved):
LGTM

PR #18, opened 2021-10-28
Assembly with Megahit

erikrikarddaniel:
## PR checklist

I added Megahit assembly in a rather simplistic way: Local module, only interleaved sequences.


conf/modules.config
```diff
@@ -29,8 +29,17 @@ params {
             args            = '--fastqc'
             publish_files   = ['txt':'', 'html':'fastqc', 'zip':'fastqc']
         }
+        'megahit' {
+            args            = ''
+        }
         'multiqc' {
```
emnilsson, line 35: Should both `megahit` and `megahit_interleaved` be included in the configs?

modules/local/megahit/interleaved.nf
```diff
@@ -0,0 +1,54 @@
...
+    path "versions.yml"                                            , emit: versions
+
+    script:
+    
+    """
+    megahit \\
+        --12 ${intl_pe_reads.join(',')} \\
+        -t $task.cpus \\
+        $options.args \\
+        --out-prefix $assembly
```
emnilsson, line 40: Does this mean that the module will only be able to handle interleaved pair-end input? Or would it also work on single-end reads?

modules/nf-core/modules/seqtk/mergepe/main.nf
```diff
@@ -0,0 +1,52 @@
...
+
+    output:
+    tuple val(meta), path("*.fastq.gz"), emit: reads
+    path "versions.yml"          , emit: versions
+
+    script:
+    def prefix = options.suffix ? "${meta.id}${options.suffix}" : "${meta.id}"
+    if (meta.single_end) {
+        """
+        cp ${reads} ${prefix}.fastq.gz
```
emnilsson, line 32: Would it be possible to update the version of `seqtk/mergepe` that is used? It is updated so that it doesn't copy but instead symlinks the file.

emnilsson (approved):
I have a few questions/comments, but in general I think this does look good and that it can be merged!

modules/local/megahit/interleaved.nf
erikrikarddaniel: At the moment only interleaved pe. If we at some point want to include se, we can add that later. (I think `khmer` will output some se reads in addition to pe that could go into the assembly. I chose to ignore this in the previous pipeline.)

conf/modules.config
erikrikarddaniel: I chose to have just one config option. Unless we at some point need to differentiate between options for two different modules, let's do that then. Right now I think I reasoned that calling it "megahit" was easiest for an end user.

PR #22, opened 2021-11-04
Diginorm

erikrikarddaniel:
First version of digital normalization, basically doing the first two steps of the "recipe" here: https://khmer-protocols.readthedocs.io/en/latest/mrnaseq/2-diginorm.html.
The instructions are a bit opaque, but I think these two steps will be enough, but might not use all the possible data.
We will need to run it on a larger dataset to see that it works properly.


emnilsson (approved):
LGTM!

PR #30, opened 2021-12-17
Add prodigal module (already tested)

danilodileo:
## PR checklist

- [ ] I added prodigal module and insert it inside metatdenovo.nf script.

### DIFF
--- a/assets/samplesheet.csv
+++ b/assets/samplesheet.csv
@@ -1,3 +1,3 @@
 sample,fastq_1,fastq_2
-SAMPLE_PAIRED_END,/path/to/fastq/files/AEG588A1_S1_L002_R1_001.fastq.gz,/path/to/fastq/files/AEG588A1_S1_L002_R2_001.fastq.gz
-SAMPLE_SINGLE_END,/path/to/fastq/files/AEG588A4_S4_L003_R1_001.fastq.gz,
+test_minigut,https://github.com/nf-core/test-datasets/raw/mag/test_data/test_minigut_R1.fastq.gz,https://github.com/nf-core/test-datasets/raw/mag/test_data/test_minigut_R2.fastq.gz
+test_minigut_sample2,https://github.com/nf-core/test-datasets/raw/mag/test_data/test_minigut_sample2_R1.fastq.gz,https://github.com/nf-core/test-datasets/raw/mag/test_data/test_minigut_sample2_R2.fastq.gz
--- a/conf/modules.config
+++ b/conf/modules.config
@@ -60,6 +60,9 @@ params {
         }
         'prodigal' {
             args                = ''
+            publish_dir     = "Prodigal"
+            output_format   = "gff"
+            publish_by_meta = ['assembler', 'id']
         }
         'prokka' {
             args            = ''
--- a/conf/modules.config
+++ b/conf/modules.config
@@ -60,9 +60,12 @@ params {
         }
         'prodigal' {
             args                = ''
+            publish_dir         = "Prodigal"
+            output_format       = "gff"
+            publish_by_meta     = ['assembler', 'id']
         }
         'prokka' {
-            args            = ''
+            args                = ''
         }
         'multiqc' {
             args                = ''
@@ -70,5 +73,9 @@ params {
         'seqtk_mergepe' {
             args                = ''
         }
+        'unpigz' {
+            args                = ''
+        }
     }
 }
+
--- a/conf/modules.config
+++ b/conf/modules.config
@@ -59,7 +59,7 @@ params {
             publish_files       = ['.bam':'', '.log':'log']
         }
         'prodigal' {
-            args                = ''
+            args                = '-p meta'
             publish_dir         = "Prodigal"
             output_format       = "gff"
             publish_by_meta     = ['assembler', 'id']
@@ -73,8 +73,9 @@ params {
         'seqtk_mergepe' {
             args                = ''
         }
-        'unpigz' {
+        'unpigz_megahit_contigs' {
             args                = ''
+            publish_files       = false
         }
     }
 }
--- a/conf/modules.config
+++ b/conf/modules.config
@@ -60,9 +60,7 @@ params {
         }
         'prodigal' {
             args                = '-p meta'
-            publish_dir         = "Prodigal"
-            output_format       = "gff"
-            publish_by_meta     = ['assembler', 'id']
+            publish_dir         = "prodigal"
         }
         'prokka' {
             args                = ''
--- a/conf/modules.config
+++ b/conf/modules.config
@@ -59,17 +59,13 @@ params {
             publish_files       = ['.bam':'', '.log':'log']
         }
         'prodigal' {
-<<<<<<< HEAD
             args                = '-p meta'
             publish_dir         = "prodigal"
-=======
-            args                = ''
-            publish_dir     = "Prodigal"
-            output_format   = "gff"
-            publish_by_meta = ['assembler', 'id']
->>>>>>> b85e0bbc45a84d7cf5a8531177e4da77e7ec4247
         }
         'prokka' {
+            args                = ''
+        }
+        'featurescounts' {
             args                = ''
         }
         'multiqc' {
--- a/nextflow.config
+++ b/nextflow.config
@@ -38,7 +38,10 @@ params {
     multiqc_config              = null
     multiqc_title               = null
     max_multiqc_email_size      = '25.MB'
-
+   
+    // ORF caller parameters
+    orf_caller               = 'prodigal'
+    
     // Boilerplate options
     outdir                      = './results'
     tracedir                    = "${params.outdir}/pipeline_info"
--- a/nextflow.config
+++ b/nextflow.config
@@ -19,8 +19,8 @@ params {
     diginorm                    = false
     diginorm_C                  = 20
     diginorm_k                  = 20
-    orf_caller                  = null
-
+    //orf_caller                  = null
+    
     // Trimming
     clip_r1                     = null
     clip_r2                     = null
--- a/nextflow.config
+++ b/nextflow.config
@@ -19,7 +19,7 @@ params {
     diginorm                    = false
     diginorm_C                  = 20
     diginorm_k                  = 20
-    //orf_caller                  = null
+    orf_caller                  = null
     
     // Trimming
     clip_r1                     = null
@@ -38,9 +38,6 @@ params {
     multiqc_config              = null
     multiqc_title               = null
     max_multiqc_email_size      = '25.MB'
-   
-    // ORF caller parameters
-    orf_caller                  = 'prodigal'
     
     // Boilerplate options
     outdir                      = './results'
--- a/nextflow.config
+++ b/nextflow.config
@@ -20,7 +20,7 @@ params {
     diginorm_C                  = 20
     diginorm_k                  = 20
     orf_caller                  = null
-    training_file               = '-t $params.prodigal_trainingfile'
+    prodigal_trainingfile      = null
 
     // Trimming
     clip_r1                     = null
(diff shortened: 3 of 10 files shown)


conf/modules.config
```diff
@@ -60,6 +60,9 @@ params {
         }
         'prodigal' {
             args                = ''
+            publish_dir     = "Prodigal"
```
emnilsson: A minor thing, but maybe it should be `prodigal` (with a lower case 'p') and not `Prodigal`?

conf/modules.config
```diff
@@ -60,6 +60,9 @@ params {
...
+            output_format   = "gff"
```
emnilsson: We will also need the `faa` files, for e.g. eggnog and taxonomy downstream, and shouldn't it be `publish_files` and not `output_format`? (Not sure though.)

workflows/metatdenovo.nf
```diff
@@ -179,11 +179,14 @@ workflow METATDENOVO {
...
 
-//    //
-//    // MODULE: Call Prodigal
-//    //
-//    PRODIGAL([ [id: 'full_assembly' ], MEGAHIT_INTERLEAVED.out.contigs)
-//    ch_versions = ch_versions.mix(PRODIGAL.out.versions)
+    //
+    // MODULE: Call Prodigal
+    //
+    PRODIGAL(
```
emnilsson: Is it intentional that you haven't implemented any `if`-statement for deciding which `--orf_caller` to use? Because it will be needed sooner or later.

emnilsson (changes requested):
This is a really good start, but there's somethings that should be fixed before you can merge this (see specific comments). Also, the PR was made against the `master`-branch and you should make your PR's against `dev` so I moved it there (it's possible to click edit and just change, so very easy). And we still need to implement some kind of check for which `--orf_caller` to use, and probably decide on one that should be default as well.

danilodileo:
I added the If statement and the params.orf_caller. I checked it and it works. 
Also, I added a new line for samplesheets.csv

assets/samplesheet.csv
```diff
@@ -1,3 +1,3 @@
 sample,fastq_1,fastq_2
```
erikrikarddaniel, line 1: I wonder what the function of this file is?
Test data should be pointed to from the respective `conf/test.config` file and point to the url in `test-datasets` where it can be found.

workflows/metatdenovo.nf
```diff
@@ -73,6 +79,10 @@ include { DIGINORM } from '../subworkflows/local/diginorm' addParams(
 def multiqc_options   = modules['multiqc']
 multiqc_options.args += params.multiqc_title ? Utils.joinModuleArgs(["--title \"$params.multiqc_title\""]) : ''
 
+params.prodigal_trainingfile = file("../test_prodigal/training_file.trn")
```
erikrikarddaniel: This doesn't belong here, should be set in the relevant config file, e.g. `conf/test.config`.

workflows/metatdenovo.nf
```diff
@@ -179,12 +189,26 @@ workflow METATDENOVO {
...
+    if(params.orf_caller == 'prodigal'){
+        PRODIGAL(
+        UNPIGZ_MEGAHIT_CONTIGS.out.unzipped.collect { [ [ id: 'all_samples' ], it ] },
+        //[ [ id: 'all_samples' ], UNPIGZ_MEGAHIT_CONTIGS.out.test ],
+        'gff'
+    )
+    ch_prodigal_gff = PRODIGAL.out.gene_annotations
+    ch_prodigal_aa  = PRODIGAL.out.amino_acid_fasta
+    ch_prodigal_fna = PRODIGAL.out.nucleotide_fasta
+    ch_versions = ch_versions.mix(PRODIGAL.out.versions)
```
erikrikarddaniel: Fix indentation.

workflows/metatdenovo.nf
```diff
@@ -179,12 +189,26 @@ workflow METATDENOVO {
...
-
+    //
+    // MODULE: Call Prodigal
+    //
+    UNPIGZ_MEGAHIT_CONTIGS(ch_assembly_contigs)
+    //ch_assembly_contigs_unzipped = UNPIGZ_MEGAHIT_CONTIGS.out.gunzip
+    ch_versions = ch_versions.mix(UNPIGZ_MEGAHIT_CONTIGS.out.versions)
+    
+    ch_prodigal = Channel.empty()
+    if(params.orf_caller == 'prodigal'){
```
erikrikarddaniel: So, constants were not used in the example code?

erikrikarddaniel (commented):
Didn't you update the schema with the new parameters (both `orf_caller` and the one you used for training data)?

assets/samplesheet.csv
Danilo2771: > 

I didn't, how should I do?

danilodileo:
> Didn't you update the schema with the new parameters (both `orf_caller` and the one you used for training data)?

I didn't, how should I do?

assets/samplesheet.csv
danilodileo: I just change them based on the new test files but probably it doesen't matter unless you specify with --input

erikrikarddaniel:
> > Didn't you update the schema with the new parameters (both `orf_caller` and the one you used for training data)?
> 
> I didn't, how should I do?

There's a tool in `nf-core` that helps you: `nf-core schema build` (remember to change into the correct branch).
More documentation here: https://nf-co.re/developers/adding_pipelines#nf-core-pipeline-structure. (One is not advised to update it with a text editor, but if you're working in sequoia, it will be difficult with the web interface. Either do it with a text editor, copying a relevant example param, or push the branch and pull to your local computer and do the update there.)

workflows/metatdenovo.nf
danilodileo: The module works without constants. It is easier to write.

assets/samplesheet.csv
erikrikarddaniel: I have no idea. Make sure you update `conf/test.config` with the new `samples.csv` (use the url).

workflows/metatdenovo.nf
erikrikarddaniel: It's not the module that should have a constant, it's the if clause that should maybe look like this:
```
if ( params.orf_caller == ORF_CALLER_PRODIGAL ) {
```
With `ORF_CALLER_PRODIGAL` declared somewhere. This is good practice in most programming languages, but we can live without that for the moment.

workflows/metatdenovo.nf
```diff
@@ -179,12 +188,26 @@ workflow METATDENOVO {
...
+    UNPIGZ_MEGAHIT_CONTIGS(ch_assembly_contigs)
+    //ch_assembly_contigs_unzipped = UNPIGZ_MEGAHIT_CONTIGS.out.gunzip
+    ch_versions = ch_versions.mix(UNPIGZ_MEGAHIT_CONTIGS.out.versions)
+    
+    ch_prodigal = Channel.empty()
+    if( params.orf_caller == 'prodigal' ) {
+        PRODIGAL(
+            UNPIGZ_MEGAHIT_CONTIGS.out.unzipped.collect { [ [ id: 'all_samples' ], it ] },
+            //[ [ id: 'all_samples' ], UNPIGZ_MEGAHIT_CONTIGS.out.test ],
+            'gff'
```
erikrikarddaniel: Also this should be indented (now I did it).

workflows/metatdenovo.nf
```diff
@@ -179,12 +188,26 @@ workflow METATDENOVO {
...
+    //
+    UNPIGZ_MEGAHIT_CONTIGS(ch_assembly_contigs)
+    //ch_assembly_contigs_unzipped = UNPIGZ_MEGAHIT_CONTIGS.out.gunzip
+    ch_versions = ch_versions.mix(UNPIGZ_MEGAHIT_CONTIGS.out.versions)
+    
+    ch_prodigal = Channel.empty()
+    if( params.orf_caller == 'prodigal' ) {
+        PRODIGAL(
+            UNPIGZ_MEGAHIT_CONTIGS.out.unzipped.collect { [ [ id: 'all_samples' ], it ] },
+            //[ [ id: 'all_samples' ], UNPIGZ_MEGAHIT_CONTIGS.out.test ],
```
erikrikarddaniel: Remove lines like this that are just remnants of old attempts.

erikrikarddaniel (commented):
Looking better, but schema still not updated...

workflows/metatdenovo.nf
danilodileo: I changed this line so now it has ORF_CALLER_PRODIGAL in it

nextflow.config
```diff
@@ -20,7 +20,7 @@ params {
     diginorm_C                  = 20
     diginorm_k                  = 20
     orf_caller                  = null
-
+
```
erikrikarddaniel: I think this (space characters at the end of the line) will fail in linting, so remove the spaces. Also below, line 41.

nextflow_schema.json
```diff
@@ -81,7 +81,8 @@
                 },
                 "orf_caller": {
                     "type": "string",
-                    "description": "Specify which ORF caller you would like to run."
+                    "description": "Specify which ORF caller you would like to run.",
```
erikrikarddaniel: Add a description of possible alternatives, e.g. ", possible alternatives: prodigal, default: prodigal.". In the near future the list will be added to.
We might also want a fuller description, which goes after "help_text" IIRC.

nextflow.config
```diff
@@ -20,7 +20,7 @@ params {
     diginorm_C                  = 20
```
erikrikarddaniel, line 20: I'm missing the prodigal training file parameter here and in `nextflow_schema.json`.

erikrikarddaniel (changes requested):
The new parameter for prodigal training file is missing from `nextflow.config` and `nextflow_schema.json`. In the latter it should be marked as "hidden".
A few smaller changes here and there too.

nextflow_schema.json
danilodileo: I changed it!

nextflow.config
danilodileo: I added the training file parameter but I don't know If I did in the right way

emnilsson (approved):

nextflow_schema.json
```diff
@@ -81,7 +81,14 @@
                 },
                 "orf_caller": {
                     "type": "string",
-                    "description": "Specify which ORF caller you would like to run."
+                    "description": "possible alternatives: prodigal, default: prodigal.",
```
erikrikarddaniel: Well, you should have kept the first part, to make it "Specify which ORF caller you would like to run, possible alternatives: prodigal, default: prodigal."!

nextflow_schema.json
```diff
@@ -81,7 +81,14 @@
...
+                    "default": "prodigal"
+                },
+                "training_file": {
+                    "type": "string",
+                    "default": "-t $params.prodigal_trainingfile",
+                    "hidden": true,
+                    "description": "add a training file to prodigal"
                 }
```
erikrikarddaniel: This chunk should define `prodigal_trainingfile`. No need for a default, the user should specify a path if he/she wants to use one, otherwise this is not set. (The default is hence null.)

nextflow.config
```diff
@@ -20,6 +20,7 @@ params {
...
     diginorm_k                  = 20
     orf_caller                  = null
+    training_file               = '-t $params.prodigal_trainingfile'
```
erikrikarddaniel: This should be null, so it's not added if the user doesn't specify it.

erikrikarddaniel (changes requested):
A few more comments!

erikrikarddaniel (approved):
LGTM.

PR #33, opened 2022-01-07
Add prokka subworkflow and make sure it runs as an option to prodigal

### LINKED ISSUE #6: Annotate contigs with Prokka
Prokka tends to fail on large sets of sequences, so subset them to smaller sets. Check out https://www.nextflow.io/docs/latest/operator.html#splitfasta for splitting the assembly.

emnilsson:
Added `prokka` as a subworkflow together with the `cat/cat` module so that the output is concatenated and the desired files are copied to the results directory. However, that doesn't work since I updated which `cat/cat` module is used, I haven't been able to find out why but I suspect it might have to do with the fact that arguments are given differently with the new modules-syntax. Hopefully this will be resolved when we perform the template-sync.
Another thing is that linting ends up in failed tests, they are mostly regarding files not existing or not matching the template, something we will need to fix in the future.

This resolves #6 `Annotate contigs with PROKKA`.

## PR checklist

- [-] Make sure your code lints (`nf-core lint`).
- [x] Ensure the test suite passes (`nextflow run . -profile test,docker`).


conf/modules.config
```diff
@@ -64,6 +64,12 @@ params {
         }
         'prokka' {
             args                = ''
+            publish_files       = false
+        }
+        'cat_cat' {
```
erikrikarddaniel: Perhaps this should have another name, one that is more transparent and shows that it's called for Prokka output -- `prokka_cat` maybe?

subworkflows/local/prokka_cat.nf
```diff
@@ -0,0 +1,34 @@
...
+// Run PROKKA on contigs that are split by size, then concatenate output and gunzip it
+//
+
+params.prokka_options    = [:]
+params.cat_options       = [:]
+
+include { PROKKA               } from '../../modules/nf-core/modules/prokka/main'           addParams( options: params.prokka_options     )
+include { CAT_CAT as GFF_CAT   } from '../../modules/nf-core/modules/cat/cat/main'          addParams( options: params.cat_options        )
+include { CAT_CAT as FAA_CAT   } from '../../modules/nf-core/modules/cat/cat/main'          addParams( options: params.cat_options        )
+include { CAT_CAT as FNA_CAT   } from '../../modules/nf-core/modules/cat/cat/main'          addParams( options: params.cat_options        )
```
erikrikarddaniel, line 11: If one has different entries for these in `modules.config`, one _can_, but doesn't have to, set params differently.
(I'm not completely up to date on subworkflow syntax though.)

workflows/metatdenovo.nf
```diff
@@ -11,9 +11,10 @@ WorkflowMetatdenovo.initialise(params, log)
 
 // Validate parameters Orfcaller:
 def valid_params = [
-    orf_caller  : ['prodigal']
+    orf_caller  : ['prodigal','prokka']
 ]
 ORF_CALLER_PRODIGAL = 'prodigal'
+ORF_CALLER_PROKKA   = 'prokka'
```
erikrikarddaniel: If we use constants, like these, they should be used in the definition of `valid_params` too, i.e.:
```
def valid_params = [
    orf_caller = [ORF_CALLER_PRODIGAL, ORF_CALLER_PROKKA]
]
```
(The definitions of the constants need to come before this then.)
I'm not sure this is common practice in nf-core.

workflows/metatdenovo.nf
```diff
@@ -71,6 +72,17 @@ include { DIGINORM } from '../subworkflows/local/diginorm' addParams(
     diginorm_extractpairedreads_options: diginorm_extractpairedreads_options
 )
 
+//
+// SUBWORKFLOW: Consisting of nf-core/modules
+//
+def prokka_options              = modules['prokka']
+def cat_options                 = modules['cat_cat']
```
erikrikarddaniel: Could be one each for `gff_cat` etc. Maybe better to just rename the `modules.config` entry to `prokka_cat`?

erikrikarddaniel (approved):
Looks fine. Just one small thing: Naming of entry in `modules.config`, see comments in code.

subworkflows/local/prokka_cat.nf
emnilsson: Sure, but I'm having a hard time seeing a use case when you want different params for them? I kind of liked having the same since that saves a lot of words..

conf/modules.config
emnilsson: Yes, seems logical! At first I thought the name needed to be the same as the actual module, but understood later that it wasn't the case.

workflows/metatdenovo.nf
emnilsson: I haven't found any examples in the nf-core pipelines I've looked at, so thanks for the comment, will fix.

PR #39, opened 2022-02-02
Transdecoder

danilodileo:
## PR checklist

- [x] added two more modules (transdecoder/predict, transdecoder/longorf)
- [x] created a new sub-workflow TRANSDECODER
- [x] modified metatdenovo.nf
- [x] pipeline tested with test files


workflows/metatdenovo.nf
```diff
@@ -212,6 +213,18 @@ workflow METATDENOVO {
         ch_versions.unique().collectFile(name: 'collated_versions.yml')
     )
     
+    //
+    // MODULE: Transdecoder Longorf
```
emnilsson: I think it's better to refer to the Subworkflow here, and maybe a few words that it's for ORF-calling?

workflows/metatdenovo.nf
```diff
@@ -212,6 +213,18 @@ workflow METATDENOVO {
...
+    //
+    
+    ch_transdecoder_longorf = Channel.empty()
+    if( params.orf_caller == ORF_CALLER_TRANSDECODER ) {
+        TRANSDECODER(
+            UNPIGZ_MEGAHIT_CONTIGS.out.unzipped.collect { [ [ id: 'all_samples' ], it ] }
```
emnilsson, line 223: Since you've created a subworkflow for this process, I think it would be good if you move the `unpigz`-part to the subworkflow - we do not want the megahit-contigs to be unzipped in case that's not needed (at least, that's my opinion).

emnilsson (changes requested):
In general, this looks good, but I also have a few general comments:
I think you need to specify which files we want to publish and where to publish them in `conf/modules.config`.
If we want to continue with tests for each tool you will need to add a test-profile for `transdecoder` (basically the same as e.g. `test_prokka.config`, which also need to be included in `nextflow.config`.)
You should add `transdecoder` as an option to the `orf_caller`'s in `nextflow_schema.json`.

workflows/metatdenovo.nf
danilodileo: I understand. Do you think that it is better to do it also in `prodigal`?  it has the same input as `transdecoder`

danilodileo:
> In general, this looks good, but I also have a few general comments:
> I think you need to specify which files we want to publish and where to publish them in `conf/modules.config`.
> If we want to continue with tests for each tool you will need to add a test-profile for `transdecoder` (basically the same as e.g. `test_prokka.config`, which also need to be included in `nextflow.config`.)
> You should add `transdecoder` as an option to the `orf_caller`'s in `nextflow_schema.json`.

done!

workflows/metatdenovo.nf
emnilsson: Yes, I think we'll need to change that, but that's something for a later PR.

emnilsson (approved):
We will need to put the `unpigz` in the processes that needs it at some point (now is a good time, but later on is fine as well), but otherwise I think this looks good and you could merge it.

PR #40, opened 2022-02-03
fixed bug transdecoder

danilodileo:
- I rewrite the `module.config` file on TRANSDECODER
- I changed the output from the org callers to work with `featurecounts`

### DIFF
--- a/conf/modules.config
+++ b/conf/modules.config
@@ -151,12 +151,21 @@ process {
         ]
     }
 
-    withName: TRANSDECODER {
+    withName: LONGORF {
         publishDir = [
-            path: { "${params.outdir}/transdecoder" }
-            mode: 'copy'
+            path: { "${params.outdir}/transdecoder" },
+            mode: 'copy',
             pattern: "*"
-
+        ]
+    }
+    
+    withName: PREDICT {
+        publishDir = [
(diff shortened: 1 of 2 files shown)


conf/modules.config
```diff
@@ -151,12 +151,21 @@ process {
         ]
     }
 
-    withName: TRANSDECODER {
+    withName: LONGORF {
```
emnilsson: Do we really need to publish files both for `longorf` and for `predict`? I think it would be tidier if we just kept the output form the process that actually produces the orf's we want to continue processing.

emnilsson (commented):
I think it looks good except the potential double-publishing of files for `LONGORF` and `PREDICT`.

conf/modules.config
danilodileo: so we can keep `predict`, right? I wasn't sure so I decided to put both :)

emnilsson: If `predict` is the one to produce the `gff` and `faa` and other files we want for downstream, then yes, let's keep that one!

danilodileo: It is! I will push the update in few minutes!

danilodileo: Done!

emnilsson (approved):
LGTM

PR #44, opened 2022-03-03
Eukulele

danilodileo:
## PR checklist


conf/modules.config
```diff
@@ -175,4 +175,15 @@ process {
             pattern: "*.gz"
         ]
     }
+    withName: EUKULELE {
+        publishDir = [
+            path: { "${params.outdir}/eukulele" },
+            mode: 'copy',
+            pattern: '*'
+            ]
```
erikrikarddaniel: I don't know what the nf-core standard is, but I would prefer if you pulled this bracket in one level, i.e. remove four spaces. That would make it similar to line 188.

conf/test_eukulele.config
```diff
@@ -0,0 +1,27 @@
...
+
+    // Limit resources so that this can run on GitHub Actions
+    max_cpus   = 4
+    max_memory = 16.GB
+    max_time   = 6.h
+
+    // Input data
+    input  = 'https://raw.githubusercontent.com/nf-core/test-datasets/magmap/samplesheets/samplesheet.csv'
+
+    eukulele_db  = '/home/dadiaa/phylodb/'
```
erikrikarddaniel: You can't have a local full path in the test config. The way to go is to specify this on the command line, i.e. `--eukulele_db`.

modules/local/eukulele/main.nf
```diff
@@ -0,0 +1,46 @@
...
+    
+    EUKulele \\
+        $args \\
+        -m mets \\
+        -o $prefix \\
+        --CPUs ${task.cpus} \\
+        -s \\
+        $contigs_fasta || rc=\$? 
+    
+    echo "EUKulele done, rc: \$rc"
```
erikrikarddaniel: Remove, this was from your test runs, I suppose.

modules/local/eukulele/main.nf
```diff
@@ -0,0 +1,46 @@
...
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    
+    """
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        eukulele: \$(echo \$(EUKulele --version 2>&1) | sed 's/Running EUKulele with command line arguments, as no valid configuration file was provided.//; s/The current EUKulele version is//g')
+    END_VERSIONS
```
erikrikarddaniel: I'm not sure if it's a good idea to have this before calling the program. I can't see that it hurts, but it's not the way it's usually done.
I think you could have it before the if clause on line 40.

modules/local/eukulele/meta.yml
```diff
@@ -0,0 +1,51 @@
+name: eukulele
+description: EUKulele is a tool for taxonomical annotation in eukaryotes
+keywords:
+  - eukaryotes
+  - tax annotation
+tools:
+  - eukulele:
+      description: Easy taxonomic annotation for eukaryotic microbes
```
erikrikarddaniel: I know you took this description from Eukulele's documentation, but since we're using it also for prokaryotes it's a bit misleading, so remove `eukaryotic`. Also on line 2 above.

modules/local/eukulele/main.nf
```diff
@@ -0,0 +1,46 @@
...
+    
+    input:
+    tuple val(meta), path(contigs_fasta)
+
+    output:
+    tuple val(meta), path("${meta.id}/taxonomy_estimation/*.out")                  , emit: taxonomy_extimation
+    tuple val(meta), path("${meta.id}/taxonomy_counts/${meta.id}_all_*_counts.csv"), emit: taxonomy_counts
+    tuple val(meta), path("${meta.id}/mets_full/diamond/*")                        , emit: diamond
+    
+    path "versions.yml"                                                          , emit: versions
```
erikrikarddaniel: Put the comma in line with the ones above.

nextflow.config
```diff
@@ -21,6 +21,8 @@ params {
     diginorm_k                  = 20
     orf_caller                  = 'prodigal'
     prodigal_trainingfile       = null
+    eukulele_db                 = null
```
erikrikarddaniel: We actually need two parameters for the eukulele database: One for which to choose (e.g. PhyloDb) and one for a path where it can be found, should be put. I think `eukulele_db` is natural for the former, and maybe `eukulele_dbpath` for the latter. The latter could have a default, e.g. `eukulele` which would make eukulele download to a directory in the same directory as the pipeline is run from. The `eukulele_db` should have a list of allowed values corresponding to what the program expects, and the pipeline should check that an allowed value was provided; similar to how the ORF caller is specified.

workflows/metatdenovo.nf
```diff
@@ -236,7 +236,17 @@ workflow METATDENOVO {
...
+    
+    //
+    // SUBWORKFLOW: Eukulele
+    //
 
+    ch_unpigz_contigs = UNPIGZ_MEGAHIT_CONTIGS.out.unzipped.collect { [ [ id: 'all_samples' ], it ] }
+    if( params.skip_eukulele) {} 
+    else {
+        SUB_EUKULELE(ch_unpigz_contigs)
+        }
```
erikrikarddaniel: The way to go is: 
```
if ( ! params.skip_eukulele ) {
    SUB_EUKULELE(ch_unpigz_contigs)
}
```
You might have to have an `else` in the future, but let's worry about that then.

modules/local/mvdir.nf
```diff
@@ -0,0 +1,13 @@
+process STAGE_FASTA_DIR {
```
erikrikarddaniel: Why does the process have a different name than the file name?

workflows/metatdenovo.nf
```diff
@@ -68,7 +68,7 @@ include { DIGINORM } from '../subworkflows/local/diginorm'
 
 include { PROKKA_CAT   } from '../subworkflows/local/prokka_cat'
 include { TRANSDECODER } from '../subworkflows/local/transdecoder'
-
+include { SUB_EUKULELE } from '../subworkflows/local/folder_eu'
```
erikrikarddaniel: Why is the subworkflow called `folder_eu`? And why can't I see that file here? Did you forget to include it in the commit?

erikrikarddaniel (changes requested):
Quite a number of things, I'm afraid.
Most important is that we need to parameters for the eukulele db, one for the name of it, another for a path to it. See comment in the config file code.
Second, it seems you didn't include all necessary files; I'm missing the subworkflow file (and it has a misleading name).

conf/test_eukulele.config
danilodileo: right, it makes sense. I will change it!

nextflow.config
```diff
@@ -21,8 +21,9 @@ params {
     diginorm_k                  = 20
     orf_caller                  = 'prodigal'
     prodigal_trainingfile       = null
-    eukulele_db                 = null
+    eukulele_db                 = 'phylodb'
     skip_eukulele               = false
+    eukulele_dbpath             = null
```
erikrikarddaniel: You got me wrong here, I think. I think we should have a default for the `eukulele_dbpath`, because otherwise the large database will just be put into a work directory. We could set the default for the path to e.g. `eukulele_db`.
I think having `phylodb` as default for the `eukulele_db` makes sense as prokaryotes will likely be the most common usecase. One could argue for not having a default at all, to ensure that the user actively chooses a database after considering the available options. In the end, I'm happy with having a `phylodb` default.

workflows/metatdenovo.nf
```diff
@@ -242,10 +249,21 @@ workflow METATDENOVO {
...
+        }
+        else if( params.eukulele_db == EUKULELE_DB_MMETSP){
+            SUB_EUKULELE(ch_unpigz_contigs)
         }
+        else if( params.eukulele_db == EUKULELE_DB_EUKPROT){
+            SUB_EUKULELE(ch_unpigz_contigs)
+        }
+        else if( params.eukulele_db == EUKULELE_DB_EUKZOO){
+            SUB_EUKULELE(ch_unpigz_contigs)
+        }
```
erikrikarddaniel: You don't need these `if`s as they all result in the same call. (I probably confused you by referring to the ORF caller.)
What we need from these values is only the code in line 25 that means whatever the user provided as value to `eukulele_db` will be checked.

erikrikarddaniel (changes requested):
OK, two to go! :-)

nextflow.config
danilodileo: Should I do something like this:
`eukulele_dbpath = '../phylodb/' ` as default?
so the database will be downloaded out of the working directory?

erikrikarddaniel: Not the two periods, just `phylodb`. Then, a directory with this name will be created and files downloaded there. A bit clearer for the end user wanting to reuse the files.

conf/modules.config
```diff
@@ -182,8 +182,9 @@ process {
             pattern: '*'
         ]
         memory = 16.GB
-        ext.args = [ 
+        ext.args = [ params.orf_caller == 'transdecoder' ? "--protein_extension .pep" : '', 
             params.eukulele_dbpath ? "--reference_dir ${params.eukulele_dbpath}" : '',
             params.eukulele_db ? "--database ${params.eukulele_db}" : ''
-        ].join(' ').trim()}
+            ].join(' ').trim()
+        }
```
erikrikarddaniel: I'm not so fond of this indentation pattern. I would prefer the `params.orf_caller ...` on line 185 on its own line and that you kept the indentation as it was on line 188. The curly bracket on line 189 should probably move one level left.

modules/local/eukulele/download.nf
```diff
@@ -0,0 +1,32 @@
+process WGET_DB {
+    tag '$meta.id'
+    label 'process_low'
```
erikrikarddaniel: Should perhaps be `process_long`? (I'm not sure how much resources the different classes take.)

modules/local/eukulele/download.nf
```diff
@@ -0,0 +1,32 @@
...
+    
+    target = params.eukulele_dbpath + "/" + params.eukulele_db + "/reference.pep.fa.gz"
+    fetchcmd = "wget -O $target $fasta_url"
+    
+    target2 = params.eukulele_dbpath + "/" + params.eukulele_db + "/taxonomy-table.txt"
+    getchcmd = "wget -O $target2 $tax_url"
+    
+    """ 
+    $fetchcmd
+    $getchcmd
```
erikrikarddaniel: Wouldn't it be clearer to build the fetch commands here instead of placing them in variables?

modules/local/eukulele/download.nf
```diff
@@ -0,0 +1,32 @@
...
+    val(tax_url)
+
+    output:
+    path("~/eukulele") , emit: database
+    path "versions.yml", emit: versions
+
+    script:
+    def args = task.ext.args ?: ''
+    
+    target = params.eukulele_dbpath + "/" + params.eukulele_db + "/reference.pep.fa.gz"
```
erikrikarddaniel: I would rather call this `fastafile` and the one on line 19 `taxonomyfile` to make it clearer what they contain.

modules/nf-core/modules/transdecoder/predict/main.nf
```diff
@@ -29,7 +29,7 @@ process TRANSDECODER_PREDICT {
         -O ${prefix} \\
         -t \\
         $fasta
-
+
```
erikrikarddaniel: Trailing spaces are not allowed in nf-core!

nextflow.config
```diff
@@ -21,10 +24,12 @@ params {
...
     orf_caller                  = 'prodigal'
     prodigal_trainingfile       = null
+    
+    // eukulele
     eukulele_db                 = 'phylodb'
     skip_eukulele               = false
-    eukulele_dbpath             = false
-
+    eukulele_dbpath             = '~/eukulele'
+
```
erikrikarddaniel: More trailing spaces.

subworkflows/local/eukulele.nf
```diff
@@ -2,26 +2,47 @@
...
+
+    if( params.eukulele_dbpath == '~/eukulele') { 
+        if( params.eukulele_db == 'phylodb') {
+            input1 = "https://www.dropbox.com/s/dl/ahpqdkmjg6xekwg/phylodb_1.076.pep.fa.gz"
+            input2 = "https://www.dropbox.com/s/dl/ak76j4po9tdcbof/phylodb_1.076.taxonomy.txt"
+        } else {
+            input1 = "https://www.dropbox.com/s/4qilp6se3qf47uq/reference-pep.fa"
+            input2 = "https://www.dropbox.com/s/w1uv1j9qr63z3ac/taxonomy_table.txt"
+        }
+    }
```
erikrikarddaniel: Shouldn't these paths be in a config file? Check out https://github.com/nf-core/ampliseq/blob/dev/conf/ref_databases.config for an example for how to build a map in a config file.

subworkflows/local/eukulele.nf
```diff
@@ -2,26 +2,47 @@
...
         db
 
-
     main:
         ch_versions = Channel.empty()
-        ch_eukulele_pathdb = Channel.empty()
+        
+        if( params.eukulele_dbpath == '~/eukulele'){ 
+            WGET_DB         ( input1, input2 )
+        }
```
erikrikarddaniel: This confuses me. Shouldn't the database be downloaded if the user specifies another directory, but the files are missing? It might be that you should always download, but use `rsync` instead of `wget` to download as that by default only downloads if the upstream file is different than the target. (`rsync -avh --size-only url target`). (Renaming the download module from `WGET` to `DOWNLOAD` would make it easier as the name is not dependent on the implementation.)

subworkflows/local/eukulele.nf
```diff
@@ -2,26 +2,47 @@
...
+        
         MV_DIR          ( contigs )
-        EUKULELE        ( MV_DIR.out.contigs_dir, db )
+        
+        if( params.eukulele_dbpath == '~/eukulele') {
+            EUKULELE        ( MV_DIR.out.contigs_dir, WGET_DB.out.database )
+        } 
+        else {
+            EUKULELE (MV_DIR.out.contigs_dir, db)
+        }
```
erikrikarddaniel: Again, I'm not sure what you're trying to achieve by this test. It looks suspicious.

workflows/metatdenovo.nf
```diff
@@ -197,17 +203,21 @@ workflow METATDENOVO {
         PROKKA_CAT(MEGAHIT_INTERLEAVED.out.contigs)
         ch_versions = ch_versions.mix(PROKKA_CAT.out.versions)
         ch_gff      = PROKKA_CAT.out.gff
-        ch_eukulele = PROKKA_CAT.out.faa
+        ch_protein  = PROKKA_CAT.out.faa
+        UNPIGZ_MEGAHIT_CONTIGS(ch_protein)
```
erikrikarddaniel: Strange name on this module. It takes a *protein* channel but has `CONTIGS` in its name. Maybe it should be just `UNPIGZ_FASTA` or something? Probably something more specific, e.g. `UNPIGZ_PROKKA_FAA` (maybe without `PROKKA`), use `as` when including. I'm suspecting you're reusing a module, which I don't think is a good idea.

workflows/metatdenovo.nf
```diff
@@ -229,11 +239,13 @@ workflow METATDENOVO {
 
     ch_transdecoder_longorf = Channel.empty()
     if( params.orf_caller == ORF_CALLER_TRANSDECODER ) {
+        UNPIGZ_MEGAHIT_CONTIGS(ch_assembly_contigs)
         TRANSDECODER(
             UNPIGZ_MEGAHIT_CONTIGS.out.unzipped.collect { [ [ id: 'all_samples' ], it ] }
         )
         ch_gff      = TRANSDECODER.out.gff.map { it[1] }
         ch_eukulele = TRANSDECODER.out.pep
+        ch_eukulele.view()
```
erikrikarddaniel: Doesn't belong in the committed code; you probably forgot it.

erikrikarddaniel (commented):
Quite a number of comments again! :-)
I think the main thing is to get the logic behind the parameters to work. The tests you do for what the params are set to make me suspicious.

modules/local/eukulele/download.nf
```diff
@@ -0,0 +1,32 @@
+process WGET_DB {
+    tag '$meta.id'
+    label 'process_low'
+    
+    input:
+    val(fasta_url)
+    val(tax_url)
+
+    output:
+    path("~/eukulele") , emit: database
```
erikrikarddaniel: Change to:
```
path($fastafile), emit: eukulele_fasta
path($taxonomyfile), emit: eukulele_taxonomy
```
If you, in the next module/process, need only the directory you can either take the basename of one of the channels, or use `params.eukulele_dbpath`.

erikrikarddaniel (approved):
Looks fine, as far as I can tell. Merge if all works, but check the failing tests below!

erikrikarddaniel (approved):
Looks good AFAIK.
But now I see you have merge conflicts...

PR #45, opened 2022-04-25
Eggnog module

danilodileo:
## PR checklist

I started working on this module from the Emelie's template script from her repository, then I added some more features!


conf/test_eggnog.config
```diff
@@ -0,0 +1,25 @@
...
+
+    Use as follows:
+        nextflow run nf-core/metatdenovo -profile test_eggnog,<docker/singularity>
+
+----------------------------------------------------------------------------------------
+*/
+
+params {
+    config_profile_name        = 'Test eggnog profile'
+    config_profile_description = 'Minimal test dataset to check pipeline with eggnog function added''
```
emnilsson: There's an extra `'` at the end of this line, it needs to be removed.

conf/test_eggnog.config
```diff
@@ -0,0 +1,25 @@
...
+    // Limit resources so that this can run on GitHub Actions
+    max_cpus   = 2
+    max_memory = '6.GB'
+    max_time   = '6.h'
+
+    // Input data
+    input  = 'https://raw.githubusercontent.com/nf-core/test-datasets/metatdenovo/samplesheet/samplesheet.csv'
+    
+    eggnog_mapper = true
```
emnilsson: If I understand the parameters correctly, it should be `eggnog` and not `eggnog_mapper`.

nextflow.config
```diff
@@ -21,7 +21,19 @@ params {
     diginorm_k                  = 20
     orf_caller                  = 'prodigal'
     prodigal_trainingfile       = null
-
+    
+    // eggnog parameters
```
emnilsson, line 25: These parameters are not defined in the `nextflow.schema`, which results in warnings. Also, are all of them needed? Isn't this a bit much for a user to define?

workflows/metatdenovo.nf
```diff
@@ -190,22 +199,27 @@ workflow METATDENOVO {
         PROKKA_CAT(MEGAHIT_INTERLEAVED.out.contigs)
         ch_versions = ch_versions.mix(PROKKA_CAT.out.versions)
         ch_gff      = PROKKA_CAT.out.gff
+        ch_protein  = PROKKA_CAT.out.faa
+        UNPIGZ_FASTA_PROTEIN(ch_protein)
+        MEGAHIT_INTERLEAVED.out.contigs.collect { [ [ id: 'all_samples' ]] }
+            .combine(UNPIGZ_FASTA_PROTEIN.out.unzipped)
+            .set{ ch_aa }
     }
```
emnilsson: Wouldn't it be better to just unzip the protein-file for `eggnog`? I.e. like you've done with the contigs for `prodigal` and `transdecoder`?

nextflow.config
```diff
@@ -132,6 +144,7 @@ profiles {
     test_full           { includeConfig 'conf/test_full.config'         }
     test_prokka         { includeConfig 'conf/test_prokka.config'       }
     test_transdecoder   { includeConfig 'conf/test_transdecoder.config' }
+    test_eggnog         { includeConfig 'conf/test_eggnog.config'       }
```
emnilsson, line 145: The test doesn't work for me, at least not with just adding `test_eggnog`, is it required to add something about downloading? Because when I run it, only `eggnog_mapper` is run and not `eggnog_download`, which then crashes because `eggnog_mapper` doesn't have anything to work with.

conf/modules.config
```diff
@@ -175,4 +175,38 @@ process {
             pattern: "*.gz"
         ]
     }
+
```
emnilsson: This row contains trailing white spaces, according to github actions (EditorConfig).

conf/modules.config
```diff
@@ -175,4 +175,38 @@ process {
...
     }
+    
+    withName: EGGNOG_MAPPER {
+        publishDir = [
+            [
+                path: { "${params.outdir}/eggnog" },
+                mode: 'copy',
+                pattern: "*.emapper.hits"
+            ],
+            [
```
emnilsson: This row contains trailing white spaces, according to github actions (EditorConfig).

conf/modules.config
```diff
@@ -175,4 +175,38 @@ process {
...
+    }
+
+    withName: EGGNOG_DOWNLOAD {
+        ext.args = [
+            params.eggnog_dwnl_hmmer ? "-H -d $params.eggnog_hmmer" : '',
+            params.eggnog_dwnl_mmseq2 ? "-M" : '',
+            params.eggnog_dwnl_pfam ? "-P" : '',
+            params.eggnog_no_dmnd ? "-D" : '',
+       ].join(' ').trim()
+    }
```
emnilsson: This row contains trailing white spaces, according to github actions (EditorConfig).

conf/modules.config
```diff
@@ -175,4 +175,38 @@ process {
...
+        ].join(' ').trim()
+    }
+
+    withName: EGGNOG_DOWNLOAD {
+        ext.args = [
+            params.eggnog_dwnl_hmmer ? "-H -d $params.eggnog_hmmer" : '',
+            params.eggnog_dwnl_mmseq2 ? "-M" : '',
+            params.eggnog_dwnl_pfam ? "-P" : '',
+            params.eggnog_no_dmnd ? "-D" : '',
+       ].join(' ').trim()
```
emnilsson: This row is lacking a space (should be nx4, here it's 7 spaces, should be 8), according to github actions (EditorConfig).

nextflow.config
```diff
@@ -21,7 +21,19 @@ params {
     diginorm_k                  = 20
     orf_caller                  = 'prodigal'
     prodigal_trainingfile       = null
-
+
```
emnilsson: This row contains trailing white spaces, according to github actions (EditorConfig).

emnilsson (changes requested):
I think this looks good (especially after some small fixes), but I haven't been able to test it because the test fails for me. (See comment regarding `nextflow.config`.)

nextflow.config
danilodileo: Yeah, you are right! I think we didn't discuss which parameters are important and which not so I just put most of them. Maybe It can be easier if there is a way to give the user the ability to insert new parameters with only one command. e.g. --add_parameters "insert here". But it is important that the user can add only parameters that are also specified inside the module otherwise it will get an error.
At the moment, I can keep the code simple and think about it in a second moment (maybe it is something that we never need).

emnilsson: Maybe we can remove them for the moment? Just having "run eggnog", "here's an eggnogdb" and "here's where to put the eggnogdb", kind of?

danilodileo: Yes, sounds good. I will do that!

nextflow.config
danilodileo: It should work now. I will finish the review and then I will push a new PR :)

workflows/metatdenovo.nf
danilodileo: The thing is that ch_protein isn't a tuple so I needed to add the  id: `'all_samples'` . if you know a nicer way to do it I will change it, this is what I did but I am sure there are better options!

emnilsson: Hm, not sure I follow - when does it have to be a tuple? For `eggnog`? What I was thinking was that we could have an `if`-condition if needed just for `prokka` if `eggnog` is specified, not automatically after running `prokka`. But that might be unnecessary trouble..

nextflow.config
emnilsson: Why not include it here? So I can test it?

nextflow.config
emnilsson: Don't forget to add them in `nextflow_schema.json`

nextflow.config
danilodileo: uhm.. I don't get it. You should be able to test it because the test_eggnog is included (e.g. -profile test_eggnog works for me)

workflows/metatdenovo.nf
danilodileo: Yes, eggnog requires a tuple to work. I can add the if-condition, I will do it now.

modules/local/eggnog/download.nf
```diff
@@ -37,7 +37,8 @@ process EGGNOG_DOWNLOAD {
 
     """
 
-    touch ./eggnog/eggnog.db
+    touch ./eggnog/eggnog.db.gz
```
erikrikarddaniel: What's the `touch` for? Touching will produce an empty file if it's not present already, which will make the unzipping to fail (gzipped empty files are not empty). I think that if the `touch` is there to make sure there's a file one would have to do this with an `if` to circumvent this problem.

subworkflows/local/eggnog.nf
```diff
@@ -0,0 +1,39 @@
+//
+// Run eggnog-mapper on called ORFs, first optionally downloading the required databases
+//
+
+include { EGGNOG_DOWNLOAD } from '../../modules/local/eggnog/download'
+include { EGGNOG_MAPPER   } from '../../modules/local/eggnog/mapper'
+
+workflow EGGNOG {
+    take:
+        faa
```
erikrikarddaniel, line 10: Doesn't have to be declared as `file faa` as one does in a module? (I haven't written a subworkflow for ages.)

subworkflows/local/eggnog.nf
```diff
@@ -0,0 +1,39 @@
+//
+// Run eggnog-mapper on called ORFs, first optionally downloading the required databases
+//
+
+include { EGGNOG_DOWNLOAD } from '../../modules/local/eggnog/download'
+include { EGGNOG_MAPPER   } from '../../modules/local/eggnog/mapper'
```
erikrikarddaniel, line 6: Aren't there official modules yet? Should we contribute? Isn't anyone else using eggnog-mapper?

workflows/metatdenovo.nf
```diff
@@ -39,15 +39,18 @@ ch_multiqc_custom_config = params.multiqc_config ? Channel.fromPath(params.multi
     IMPORT LOCAL MODULES/SUBWORKFLOWS
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 */
+
 //
 // MODULE: local
 //
 include { MEGAHIT_INTERLEAVED              } from '../modules/local/megahit/interleaved.nf'
 include { UNPIGZ as UNPIGZ_MEGAHIT_CONTIGS } from '../modules/local/unpigz.nf'
+include { UNPIGZ as UNPIGZ_FASTA_PROTEIN   } from '../modules/local/unpigz.nf'
```
erikrikarddaniel: Would it perhaps be better to make sure the modules (e.g. eggnog-mapper) unzips their inputs? I think that's more inline with nf-core recommendations.

erikrikarddaniel (commented):
I'm not sure where we are with this, is it ready to approve? Does the test pass?

Small comments, several are questions more than requests for change.

subworkflows/local/eggnog.nf
danilodileo: Now that the module is working we could contribute by adding the module to nf-core. I can do it

subworkflows/local/eggnog.nf
danilodileo: It should be only with `faa`

subworkflows/local/eggnog.nf
erikrikarddaniel: It is a bit tricky because of the download of a very large database. I know that James Fellow Yates have been working on other modules that require database download, so we should check what they look like. Phyloplan is one of them, I believe.

workflows/metatdenovo.nf
danilodileo: do you mean by having a subworkflow? or having the unzip in eggnog-mapper module?

erikrikarddaniel: What I was thinking of here was to include the unzipping in the module, so you don't have to unzip beforehand. I'm not really sure what nf-core recommends.

erikrikarddaniel: Leave it as it is for now, and we can discuss later.

conf/modules.config
```diff
@@ -176,6 +176,32 @@ process {
...
+                path: { "${params.outdir}/eggnog" },
+                mode: 'copy',
+                pattern: "*.emapper.seed_orthologs"
+            ],
+            [
+                path: { "${params.outdir}/eggnog" },
+                mode: 'copy',
+                pattern: "*.emapper.annotations"
+            ]
+
```
emnilsson, line 206: It's missing an end curly bracket, }, here, might fix the issue with the tests.

emnilsson (approved):

emnilsson:
There's still something off, but not sure where the problem is, might be best to merge and then we'll sort it in the future.

PR #49, opened 2022-08-17
Hmmsearch + Hmmrank

danilodileo:
## PR checklist

I made a two new modules, hmmsearch and hmmrank 
By providing a path to a folder with .hmm profiles, hmmsearch will output *.tblout files. 
hmmrank will rank the profiles for each matched orf. The output table can be explored to check the best profile for a specific orf. 

I don't have a test as I should provide a folder with .hmm profiles. to test it on sequoia you can use the profiles in this path '/home/dadiaa/hmmrank/profiles/*.hmm' 

e.g. nextflow run . -profile test,docker --hmmsearch --hmmfilepath '/home/dadiaa/hmmrank/profiles/*.hmm'


conf/modules.config
```diff
@@ -158,6 +158,13 @@ process {
             pattern: "*"
         ]
     }
+    withName: 'HMMER' {
+        publishDir = [
+            path: { "${params.outdir}/hmmsearch" },
+            mode: 'copy',
+            pattern: "*"
+        ]
+    }
```
erikrikarddaniel: Is this necessary, or just something left? The one below `HMMER_HMMSEARCH` looks like a better match.

modules/local/hmmrank.nf
```diff
@@ -0,0 +1,78 @@
...
+
+    """
+    #!/usr/bin/env Rscript
+    library(optparse)
+    library(readr)
+    library(dplyr)
+    library(dtplyr)
+    library(tidyr)
+    library(data.table)
+    library(stringr)
```
erikrikarddaniel: Check which are actually needed by the code. I know that `optparse` is not, as it's for grabbing command line options. And I wonder if we really need to use `data.table` and `dtplyr`. They can make the code faster, but also requires a mulled container making upgrades more complicated. It would be great if we can use the `tidyverse` container as is. This means, however, that the code needs some changes.

modules/local/hmmrank.nf
```diff
@@ -0,0 +1,78 @@
...
+        ) %>%
+        transmute(accno, profile, hmm_profile = t1, evalue = as.double(evalue), score = as.double(score)) %>%
+        data.table()
+      if ( nrow(t) > 0 ) {
+        tlist[[i]] <- t
+        i <- i + 1
+      }
+    }
+    if ( length(tlist) > 0  ) {
+      tblout <- rbindlist(tlist)
```
erikrikarddaniel: This is `data.table` code, but can be replaced with `tidyverse`.

modules/local/hmmrank.nf
```diff
@@ -0,0 +1,78 @@
...
+    } else {
+      write("No results found, exiting", stderr())
+      quit(save = 'no', status = 0)
+    }
+
+    setkey(tblout, profile, accno)
```
erikrikarddaniel: Also `data.table` code.

modules/local/hmmrank.nf
```diff
@@ -0,0 +1,78 @@
...
+        rename(orf = accno) %>%
+        ungroup() %>%
+        filter(!is.na(orf)) %>%
+        as_tibble()
+
+    # write output
+    write_tsv(
+        tblout,
+        'hmmrank.out'
+        )
```
erikrikarddaniel: I think this should be pulled in one level.

workflows/metatdenovo.nf
```diff
@@ -224,8 +227,18 @@ workflow METATDENOVO {
...
         )
         ch_gff = TRANSDECODER.out.gff.map { it[1] }
+        ch_hmmr_aa = TRANSDECODER.out.pep
     }
 
+    //
+    // MODULE: Hmmsearch on orf caller output
+    //
+    if( params.hmmsearch) {
+        ch_hmmrpath = Channel.fromPath(params.hmmpath)
```
erikrikarddaniel: I assume this code only works if `hmmpath` has a value like `hmms/*.hmm`, i.e. a file glob. I think it might be better to separate `hmmpath` into `hmmdirectory` and `hmmpattern`, the latter with a default of `*.hmm`.

workflows/metatdenovo.nf
```diff
@@ -89,6 +90,7 @@ include { SUBREAD_FEATURECOUNTS as FEATURECOUNTS_CDS } from '../modules/nf-core/
 include { PRODIGAL                                   } from '../modules/nf-core/modules/prodigal/main'
 include { MULTIQC                                    } from '../modules/nf-core/modules/multiqc/main'
 include { CUSTOM_DUMPSOFTWAREVERSIONS                } from '../modules/nf-core/modules/custom/dumpsoftwareversions/main'
+include { HMMER_HMMSEARCH as HMMER                   } from '../modules/nf-core/modules/hmmer/hmmsearch/main.nf'
```
erikrikarddaniel: Better to call it `HMMSEARCH`, i.e. change to `as HMMSEARCH`.

workflows/metatdenovo.nf
```diff
@@ -224,8 +227,18 @@ workflow METATDENOVO {
...
         ch_gff = TRANSDECODER.out.gff.map { it[1] }
+        ch_hmmr_aa = TRANSDECODER.out.pep
     }
 
+    //
+    // MODULE: Hmmsearch on orf caller output
+    //
+    if( params.hmmsearch) {
+        ch_hmmrpath = Channel.fromPath(params.hmmpath)
+        HMMER(ch_hmmr_aa, ch_hmmrpath )
```
erikrikarddaniel: Change to `HMMSEARCH` here too.

erikrikarddaniel (commented):
We should have a test. The problem is how to put a directory of test files on GitHub. They had a workaround before, but I'm not sure how this is dealt with now. One could have two alternative parameters, one being a pair of `--hmmdirectory` and `--hmmpattern` (see my other comment about separating into two), the other being `--hmmfiles` containing a comma-separated list of files.

The R code can do with some clean up, but that would be easier to fix once we have a test.

conf/modules.config
danilodileo: Yes, we don't need it. I'll remove it

modules/local/hmmrank.nf
danilodileo: Ok, I'll try to write the code with tidyverse.

modules/local/hmmrank.nf
danilodileo: apparently, there isn't a function to change 'setkey' with another function in Tidyverse. any idea?

danilodileo:
> We should have a test. The problem is how to put a directory of test files on GitHub. They had a workaround before, but I'm not sure how this is dealt with now. One could have two alternative parameters, one being a pair of `--hmmdirectory` and `--hmmpattern` (see my other comment about separating into two), the other being `--hmmfiles` containing a comma-separated list of files.
> 
> The R code can do with some clean up, but that would be easier to fix once we have a test.

Can you explain me better the `--hmfiles`I made the changes and now it is possible to call separately `--hmmdirectory`and `--hmmpattern`

workflows/metatdenovo.nf
```diff
@@ -254,9 +274,12 @@ workflow METATDENOVO {
         TRANSDECODER(
             UNPIGZ_CONTIGS.out.unzipped.collect { [ [ id: 'all_samples' ], it ] }
         )
-        ch_gff = TRANSDECODER.out.gff.map { it[1] }
+        ch_gff      = TRANSDECODER.out.gff.map { it[1] }
+        ch_hmm_aa   = TRANSDECODER.out.pep
+        ch_aa       = TRANSDECODER.out.pep
+        ch_gff      = TRANSDECODER.out.gff.map { it[1] }
```
emnilsson, line 300: Why are there two ch_gff?

emnilsson (approved):
There's a lot of lint issues (trailing white spaces and some wrong number of padding), either you fix it and then merge, or you merge and someone fix it later. I assume that you tested the new R code, so go ahead and merge!

workflows/metatdenovo.nf
danilodileo: I removed one of the two ch_gff

danilodileo:
> There's a lot of lint issues (trailing white spaces and some wrong number of padding), either you fix it and then merge, or you merge and someone fix it later. I assume that you tested the new R code, so go ahead and merge!

If I use nf-core lint now, it will ask me to update all the modules. so it should be easier to merge and solve the lint issues in a second moment. I will merge now!

PR #50, opened 2022-08-19
rnaspades module

danilodileo:
Now it is possible to run RNAspades instead of MEGAHIT. the module is called with the parameters "--rnaspades", MEGAHIT is the default option.

You can try it as 'nextflow run .  -profile test, docker --rnaspades '
I need to add 'test_rnaspades.config' for this module


workflows/metatdenovo.nf
```diff
@@ -43,7 +43,7 @@ ch_multiqc_custom_config = params.multiqc_config ? Channel.fromPath(params.multi
 // MODULE: local
 //
 include { MEGAHIT_INTERLEAVED              } from '../modules/local/megahit/interleaved.nf'
-include { UNPIGZ as UNPIGZ_MEGAHIT_CONTIGS } from '../modules/local/unpigz.nf'
+include { UNPIGZ as UNPIGZ_CONTIGS } from '../modules/local/unpigz.nf'
```
erikrikarddaniel: Move the right curly brackets to be inline with each other.

workflows/metatdenovo.nf
```diff
@@ -155,15 +156,23 @@ workflow METATDENOVO {
...
+       ch_versions = ch_versions.mix(SPADES.out.versions)
+    } else {
+        MEGAHIT_INTERLEAVED(
+            ch_pe_reads_to_assembly.collect(),
+            ch_se_reads_to_assembly.collect(),
+            'all_samples'
+        )
+        ch_assembly_contigs = MEGAHIT_INTERLEAVED.out.contigs
+        ch_versions = ch_versions.mix(MEGAHIT_INTERLEAVED.out.versions)
+    }
```
erikrikarddaniel: I'd prefer a parameter `assembler` that takes "megahit" or "rnaspades" as values, with the former as default.

erikrikarddaniel (approved):
Like we have for the gene caller, having a parameter `assembler` with a value for each possibility would make the pipeline more flexible without much cost. 

Otherwise great work!

emnilsson:
I think I fixed the merge conflict, we'll notice further on otherwise, so go ahead and merge.

PR #52, opened 2022-09-20
Add files via upload

danilodileo:
## PR checklist

diagram for nf-cor/metatdenovo pipeline


emnilsson (commented):
I think the map looks good in general, but shouldn't either "intermediate step" or "Node" in the legend be green? Or what's the difference?
Also, I'd vote for a different font (some version of sans serif maybe?), but that's just a minor thing.

danilodileo:
> I think the map looks good in general, but shouldn't either "intermediate step" or "Node" in the legend be green? Or what's the difference?
> Also, I'd vote for a different font (some version of sans serif maybe?), but that's just a minor thing.

I can change the font and fix the legend. I will upload a new version of the diagram.

PR #61, opened 2022-10-20
Fix collect stats

danilodileo:
## PR checklist

Hello, 
I fixed the bug related to --skip_trimming and collect_stats.nf check the issue number 48

https://github.com/LNUc-EEMiS/metatdenovo/issues/48

### DIFF
--- a/modules/local/collect_featurecounts.nf
+++ b/modules/local/collect_featurecounts.nf
@@ -40,7 +40,7 @@ process COLLECT_FEATURECOUNTS {
                         lazy_dt() %>%
                         filter(count > 0) %>%
                         mutate(
-                            sample = str_remove(sample, 'task.sort.bam'),
+                            sample = str_remove(sample, '_T1task.sort.bam'),
                             r = count/Length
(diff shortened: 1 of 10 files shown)


modules/local/collect_featurecounts.nf
```diff
@@ -40,7 +40,7 @@ process COLLECT_FEATURECOUNTS {
                         lazy_dt() %>%
                         filter(count > 0) %>%
                         mutate(
-                            sample = str_remove(sample, 'task.sort.bam'),
+                            sample = str_remove(sample, '.sorted.bam.bam'),
```
emnilsson, line 43: Should this really be "bam.bam"? Not just ".bam"?

emnilsson (approved):
Just one comment, otherwise looks good!

modules/local/collect_featurecounts.nf
danilodileo: I checked and the output and it is .sorted.bam.bam that's why it has this str_remove

PR #62, opened 2022-11-01
Update modules

danilodileo:
## PR checklist

Hello,
I made a new PR as Dennis had some issues running the dev branch. The errors were mainly in the download of eggnog and in eukulele main.nf script. 
I ran the test for the download for both modules and now it works.

### DIFF
--- a/modules/local/eggnog/download.nf
+++ b/modules/local/eggnog/download.nf
@@ -22,6 +22,7 @@ process EGGNOG_DOWNLOAD {
     def args = task.ext.args ?: ''
 
     """
+
     mkdir eggnog
 
     download_eggnog_data.py \\
@@ -30,6 +31,24 @@ process EGGNOG_DOWNLOAD {
         --data_dir ./eggnog/
 
     ln -s ./eggnog/* ./
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        eggnog: \$( echo \$(emapper.py --version 2>&1)| sed 's/.* emapper-//' )
+    END_VERSIONS
+
+
+    """
+
+    stub:
+    
+    """
+
+    mkdir eggnog
+    touch ./eggnog/eggnog.db
+    touch ./eggnog/eggnog.taxa.db
+    touch ./eggnog/eggnog.taxa.db.traverse.pkl
+    ln -s eggnog/* ./
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
--- a/modules/local/eggnog/download.nf
+++ b/modules/local/eggnog/download.nf
@@ -41,7 +41,7 @@ process EGGNOG_DOWNLOAD {
     """
 
     stub:
-    
+ 
     """
 
     mkdir eggnog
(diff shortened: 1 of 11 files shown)


modules/local/eggnog/download.nf
```diff
@@ -26,23 +29,7 @@ process EGGNOG_DOWNLOAD {
...
         --data_dir ./eggnog/
 
-    cat <<-END_VERSIONS > versions.yml
-    "${task.process}":
-        eggnog: \$( echo \$(emapper.py --version 2>&1)| sed 's/.* emapper-//' )
-    END_VERSIONS
-
-    """
-
-    stub:
```
emnilsson, line 36: Is the stub not needed anymore? Is it possible to run the download without downloading the data then? E.g. in a test?

modules/local/eukulele/main.nf
```diff
@@ -25,7 +25,7 @@ process EUKULELE {
     """
     rc=0
     mkdir contigs
-    mv ${contigs_fasta} contigs
+    cp ${contigs_fasta} ./contigs/
```
emnilsson, line 28: Why copy and not move? And if copy, why not symlinking?

subworkflows/local/eggnog.nf
```diff
@@ -18,6 +18,7 @@ workflow EGGNOG {
         File test = new File(eggnogDB)
 
         if (! directory.exists()){
+            directory.mkdir()
```
emnilsson, line 21: I should probably run this, because I don't see how the subworkflow knows what to name the directory? Based on `eggnog_dbpath`? But isn't that an optional parameter?

emnilsson (commented):
It looks good, but I have some questions that I'd be happy if you'll answer.

subworkflows/local/eggnog.nf
danilodileo: eggnog_dbpath is "./eggnog" by default. so if the user doesn't specify any other path, the sub workflow will check whether the directory "./eggnog" (or any other name path) is available. without any directory with eggnog_dbpath name, the sub workflow will create a new directory with eggnog_dbpath.

modules/local/eukulele/main.nf
danilodileo: move will remove the contig from the work directory and it will not be available for other modules (e.g. if EUKulele starts before eggnog, the latter will not be able to use the contig file).
I tried with symlink but it is broken for some reason so I decided to use cp and it worked

modules/local/eggnog/download.nf
danilodileo: I get it now. you're right I should leave the stub there. I will put it back :)

emnilsson (commented):
Fix the stub. the rest LGTM.

modules/local/eggnog/download.nf
danilodileo: just restored the stub script

emnilsson (approved):
LGTM

PR #66, opened 2022-11-26
eukulele gtdb call

danilodileo:
## PR checklist

Hi,
I discovered that inside the nextflow pipeline, EUKulele is acting differently as outside. When I run the pipeline with gtdb path, EUKulele doesn't recognise the path and start to download a new database as default (e.g. PhyloDB). We should try do avoid this and make it work properly.
when I run EUKuele with gtdb I use this command line:

EUKulele --m mets --sample_dir contigs --reference_dir gtdb/ -o output

the EUKulele version that I use in the command line is 2.0.2 while the one in nf-core module is 2.0.3 I don't know if this affects the performance of the program.

### DIFF
--- a/modules/local/eukulele/main.nf
+++ b/modules/local/eukulele/main.nf
@@ -2,10 +2,10 @@ process EUKULELE {
     tag "$meta.id"
     label 'process_high'
 
-    conda (params.enable_conda ? "bioconda::eukulele=2.0.2-0" : null)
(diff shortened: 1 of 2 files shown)


emnilsson (approved):
LGTM! But a bit misleading name, it's basically updating the eukulele-version and unsetting the default-db for eukulele. Have we specified that phylodb is default anywhere? E.g. in the docs or in the .json? If so, we should fix that.

erikrikarddaniel (approved):
LGTM.

Issue #10, opened 2021-10-26
Nice summary tables for further analysis


erikrikarddaniel:
Summarise ORFs with taxonomic and functional annotations with counts into a set of `tsv.gz` tables.

erikrikarddaniel:
Don't forget the EGGNOG table(s). The `.annotations` table is basically tsv, but could be improved by removing the first lines that have `##` at the start plus removing the `#` starting the header line. A command like the below -- replacing the paths of course -- would do it.
```
grep -v '^##' celeribacter/eggnog/all_samples.emapper.annotations|sed 's/^#//' | gzip -c > celeribacter/summary_tables/eggnogs.tsv.gz
```

erikrikarddaniel:
Make sure keys are consistently named. I'd suggest `geneid` for the ORFs. Personally, I'd prefer all small letters in all column names (e.g. the counts table) but I'm not sure if it's worth the trouble.

erikrikarddaniel:
Make sure that tables from alternative programs, e.g. Eukulele with different databases, are named differently so they don't get overwritten after a rerun. I.e. `eukulele.phylodb.taxonomy.tsv.gz` and `eukulele.mmetsp.taxonomy.tsv.gz`.

danilodileo:
We are able to generate all the tables that we planned. I will close this issue.

PR #67, opened 2022-11-29
Eggnog table

danilodileo:
Hi, based on issue #10 I made a new table for eggnog output. Now it can be combined with counts.tsv and taxonomy.tsv table.


emnilsson (approved):
LGTM

PR #68, opened 2022-12-26
Multi eukulele

danilodileo:
I modified this PR in order to get a nicer way to call eukulele when we decide to have multiple runs.

usage:
nextflow run main.nf -profile test,docker --eukulele_db phylodb,mmetsp --eukulele_dbpath your/path/db

this update should work with the new eukulele release with --database gtdb option

N.B. this PR cover issue #10 too


erikrikarddaniel:
I think it would be better if this behaviour was triggered by `eukulele_db` being given a list, e.g. `--eukulele_db phylodb,mmetsp`.

conf/modules.config
```diff
@@ -292,10 +292,9 @@ process {
         memory = 16.GB
         ext.args = [
             params.orf_caller == 'transdecoder' ? "--protein_extension .pep" : '',
-            params.eukulele_db ? "--database ${params.eukulele_db}" : '',
             params.eukulele_method ? "-m ${params.eukulele_method}" : ''
         ].join(' ').trim()
-    }
+    ''}
```
erikrikarddaniel: Is there a newline missing?

modules/local/collect_featurecounts.nf
```diff
@@ -53,7 +53,7 @@ process COLLECT_FEATURECOUNTS {
         ) %>%
         tidyr::unnest(d) %>%
         select(-f) %>%
-        write_tsv("counts.tsv.gz")
+        write_tsv("${meta}_counts.tsv.gz")
```
erikrikarddaniel: Should be `meta.id`. You're probably passing in a `meta` that is just a name and not an object.

modules/local/eukulele/download.nf
```diff
@@ -8,19 +8,23 @@ process EUKULELE_DB {
         'quay.io/biocontainers/eukulele:2.0.3--pyh723bec7_0' }"
 
     input:
+    val(db)
 
     output:
     path "versions.yml", emit: version
-    path("phylodb/")   , emit: phylo_db , optional: true
-    path("mmetsp/")    , emit: mmetsp_db, optional: true
+    path("${db}")      , emit: db , optional: true
```
erikrikarddaniel: Should not be optional any more.

modules/local/eukulele/download.nf
```diff
@@ -8,19 +8,23 @@ process EUKULELE_DB {
...
 
     script:
     def args = task.ext.args ?: ''
 
     """
     EUKulele \\
     download \\
-    $args
+    $args \\
+    --database $db
```
erikrikarddaniel: Indent one level.

modules/local/eukulele/main.nf
```diff
@@ -8,28 +8,34 @@ process EUKULELE {
         'quay.io/biocontainers/eukulele:2.0.3--pyh723bec7_0' }"
 
     input:
-    tuple val(meta), path(contigs_fasta)
-    path(db)
+    tuple val(meta)  , path(fasta)
+    tuple path(eukdb), val(namedb)
```
erikrikarddaniel: Would it be better to have `tuple val(meta), path(fasta), val(dbname), path(eukdb)`? When you call the `EUKULELE` module with directory input (no name), pass in a `[]` for the `val(dbname)`.

subworkflows/local/eukulele.nf
```diff
@@ -2,40 +2,39 @@
...
-            } else
-                EUKULELE(fastaprot, EUKULELE_DB.out.phylo_db)
-                FORMAT_TAX(EUKULELE.out.taxonomy_estimation.map { it[1] } )
-        } else {
-                ch_database = Channel.fromPath(params.eukulele_dbpath)
-                EUKULELE(fastaprot, ch_database)
-                FORMAT_TAX(EUKULELE.out.taxonomy_estimation.map { it[1] } )
-        }
+            EUKULELE_DB(eukulele_db)
+            EUKULELE(fastaprot,EUKULELE_DB.out.db, eukulele_db)
```
erikrikarddaniel: Alternative: Combine all three channels into one and modify `EUKULELE` so it takes only one channel, see above. Requires use of `combine()` to create a channel that is the cartesian product of faa (always just one) and the eukulele databases (one or more).

modules/local/eukulele/main.nf
```diff
@@ -8,28 +8,34 @@ process EUKULELE {
...
+    gunzip   = fasta =~ /\.gz$/ ? "gunzip -c ${fasta} > ${input}" : ""
 
     """
+
+    $gunzip
+
     rc=0
     mkdir contigs
-    cp ${contigs_fasta} ./contigs/
+    cp $input ./contigs/
```
erikrikarddaniel: `mv`

erikrikarddaniel (commented):
A few comments.

conf/modules.config
danilodileo: I removed the ''

modules/local/collect_featurecounts.nf
danilodileo: solved

modules/local/eukulele/download.nf
danilodileo: solved

modules/local/eukulele/download.nf
danilodileo: done

modules/local/eukulele/main.nf
danilodileo: done

workflows/metatdenovo.nf
```diff
@@ -51,6 +52,18 @@ if ( params.hmmdir ) {
...
 
+ch_eukulele_db = Channel.empty()
+if ( !params.skip_eukulele ) {
+    if ( params.eukulele_db ) {
+        Channel
+            .of ( params.eukulele_db.split(',') )
+            .set { ch_eukulele_db }
+    } else {
+        ch_eukulele_db = []
+        }
```
danilodileo, line 63: indent

workflows/metatdenovo.nf
```diff
@@ -423,12 +418,16 @@ workflow METATDENOVO {
...
 
     if( !params.skip_eukulele){
-        // DL: I think this should also be change so that the Eukulele module does the unzipping itself.
-        if ( params.orf_caller == ORF_CALLER_PROKKA ) {
-            UNPIGZ_EUKULELE(ch_aa)
-            SUB_EUKULELE(UNPIGZ_EUKULELE.out.unzipped.collect { [ [ id: 'all_samples' ], it ] } )
-        } else
-        SUB_EUKULELE(ch_aa)
+        if ( !params.eukulele_db ) {
+            SUB_EUKULELE( ch_aa.map {[ [ id:'all_samples'], it[1], [] ] } )
```
danilodileo: add assembly.orf_caller

danilodileo:
The module has some features that needs to be fixed.
1. eukulele can't run with a custom database for taxonomic annotation
2. If you ask a double run, the sub workflow will run eukulele as many combination it can create from the .map output. for example with two database selected (--eukulele_db phylodb,mmetsp) it will try to run eukulele 4 times. 
3. it is not possible at to avoid eukulele_download module. that is an issue for the previous two points

erikrikarddaniel (approved):
LGTM.

PR #93, opened 2023-01-31
Code go thru

erikrikarddaniel:
A few small changes after we looked at the code together:

* Keep the `meta` object from the gff channels from ORF callers, and remove when needed.
* Changed `--eggnog` to `--skip_eggnog` to be consistent.
* Made sure the `meta.id` from the Prokka subworkflow is `assembly.prokka` to be in line with the other ORF callers.


emnilsson (approved):
LGTM. The tests I've run works (I've run the test profiles, except "full", "eukulele" and "eggnog", and not modified anything else). It's just a conflict that needs to be resolved, and maybe we want to change the nextflow_schema regarding the default db of eukulele?

PR #99, opened 2023-02-07
Add an eggnog summary for statistics

danilodileo:
## PR checklist

- [x] Ensure the test suite passes (`nextflow run . -profile test,docker`).

This is just one of several PRs with the goal to have a final table from `collect_stats.nf` module that will include the proportion of orf annotated with the different annotation programs. this PR is for `sum_eggnog.nf` and it is not integrated yet in `collect_stats.nf`. ( I prefer to wait and having at least the summary from `eukulele`, then I can combine all the tables in one).

### DIFF
--- a/modules/local/sum_eggnog.nf
+++ b/modules/local/sum_eggnog.nf
@@ -27,8 +27,6 @@ process SUM_EGGNOG {
     """
     #!/usr/bin/env Rscript
 
-    library(data.table)
-    library(dtplyr)
     library(dplyr)
     library(readr)
     library(tidyr)
--- a/modules/local/sum_eggnog.nf
+++ b/modules/local/sum_eggnog.nf
@@ -40,7 +40,7 @@ process SUM_EGGNOG {
 
     counts <- list.files(pattern = "*_counts.tsv.gz") %>%
         map_df(~read_tsv(.,  show_col_types  = TRUE)) %>%
-            as_tibble()
+        as_tibble()
 
     counts %>% select(1, 7) %>%
         right_join(eggnog, by = 'orf') %>%
--- a/modules/local/sum_eggnog.nf
+++ b/modules/local/sum_eggnog.nf
@@ -1,0 +1,59 @@
+process SUM_EGGNOG {
+    tag "$meta.id"
+    label 'process_low'
+
+    conda (params.enable_conda ? "conda-forge::r-tidyverse=1.3.1 conda-forge::r-data.table=1.14.0 conda-forge::r-dtplyr=1.1.0" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+    'https://depot.galaxyproject.org/singularity/mulled-v2-508c9bc5e929a77a9708902b1deca248c0c84689:0bb5bee2557136d28549f41d3faa08485e967aa1-0' :
+    'quay.io/biocontainers/mulled-v2-508c9bc5e929a77a9708902b1deca248c0c84689:0bb5bee2557136d28549f41d3faa08485e967aa1-0' } "
+
+    input:
+
+    tuple val(meta), path(fcs)), path(eggnog)
+
+    output:
+
+    path "${meta.id}_eggnog_stats.tsv", emit: overall_stats
+    path "versions.yml"               , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    """
+    #!/usr/bin/env Rscript
+
+library(data.table)
+library(dtplyr)
+library(dplyr)
+library(readr)
+library(tidyr)
+library(stringr)
+library(tidyverse)
+
+    TYPE_ORDER = c('sample', 'database', 'field', 'value')
+    # call the tables into variables
+    eggnog <- fread("eggnogs.tsv.gz")
+    counts <- list.files(pattern = "*_counts.tsv.gz") %>%
+            map_df(~fread(.))
+
+    counts[,c(1,6)] %>%
+        right_join(eggnog[,1], by = 'orf') %>%
+        drop_na() %>%
+        group_by(sample) %>%
+        count(orf) %>%
+        summarise( value = sum(n)) %>%
+        as_tibble() %>%
+        add_column(database = "eggnog", field = "n_orfs") %>%
+        relocate(value, .after = last_col()) %>%
+        write_tsv('eggnog_summary.tsv')
+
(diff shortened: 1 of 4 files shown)


modules/local/sum_eggnog.nf
```diff
@@ -0,0 +1,64 @@
...
+    library(dtplyr)
+    library(dplyr)
+    library(readr)
+    library(tidyr)
+    library(stringr)
+    library(tidyverse)
+
+    TYPE_ORDER = c('sample', 'database', 'field', 'value')
+    # call the tables into variables
+    eggnog <- read.delim("eggnogs.tsv.gz",  sep = "\t", header = TRUE, fill = TRUE) %>%
```
erikrikarddaniel: Why not `read_tsv()` or `fread()`? Both are faster. The former gives a tibble, the latter a data.table.

modules/local/sum_eggnog.nf
```diff
@@ -0,0 +1,64 @@
...
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    """
+    #!/usr/bin/env Rscript
+
+    library(data.table)
+    library(dtplyr)
```
erikrikarddaniel: I can't see that you're using neither of these two packages. Maybe it would make the code faster if you did though. If so, you need to make sure all tables are data.table and end each pipeline with a `as.data.table()`.

modules/local/sum_eggnog.nf
```diff
@@ -0,0 +1,64 @@
...
+    TYPE_ORDER = c('sample', 'database', 'field', 'value')
+    # call the tables into variables
+    eggnog <- read.delim("eggnogs.tsv.gz",  sep = "\t", header = TRUE, fill = TRUE) %>%
+        as_tibble()
+
+    counts <- list.files(pattern = "*_counts.tsv.gz") %>%
+            map_df(~read.table(.,  sep = "\t", header = TRUE, fill = TRUE)) %>%
+            as_tibble()
+
+    counts[,c(1,7)] %>%
```
erikrikarddaniel: In `dplyr` this is `counts %>% select(1, 7)`.

modules/local/sum_eggnog.nf
```diff
@@ -0,0 +1,64 @@
...
+    library(tidyverse)
+
+    TYPE_ORDER = c('sample', 'database', 'field', 'value')
+    # call the tables into variables
+    eggnog <- read.delim("eggnogs.tsv.gz",  sep = "\t", header = TRUE, fill = TRUE) %>%
+        as_tibble()
+
+    counts <- list.files(pattern = "*_counts.tsv.gz") %>%
+            map_df(~read.table(.,  sep = "\t", header = TRUE, fill = TRUE)) %>%
+            as_tibble()
```
erikrikarddaniel: I think you can use `read_tsv("*_counts.tsv.gz", show_col_types = FALSE)` here.

modules/local/sum_eggnog.nf
```diff
@@ -0,0 +1,64 @@
...
+
+    counts[,c(1,7)] %>%
+        right_join(eggnog[,1], by = 'orf') %>%
```
erikrikarddaniel: `eggnog[,1]` is hard to understand, I think. What are you actually doing?

modules/local/sum_eggnog.nf
```diff
@@ -0,0 +1,64 @@
...
+        drop_na() %>%
+        as_tibble() %>%
```
erikrikarddaniel: Hardly necessary.

modules/local/sum_eggnog.nf
```diff
@@ -0,0 +1,64 @@
...
+        group_by(sample) %>%
+        count(orf) %>%
+        summarise( value = sum(n)) %>%
+        as_tibble() %>%
```
erikrikarddaniel: Hardly necessary.

modules/local/sum_eggnog.nf
```diff
@@ -0,0 +1,64 @@
...
+            map_df(~read.table(.,  sep = "\t", header = TRUE, fill = TRUE)) %>%
+            as_tibble()
+
+    counts[,c(1,7)] %>%
+        right_join(eggnog[,1], by = 'orf') %>%
+        drop_na() %>%
+        as_tibble() %>%
+        group_by(sample) %>%
+        count(orf) %>%
+        summarise( value = sum(n)) %>%
```
erikrikarddaniel: Add a `.groups = 'drop'` as a second argument to `summarise()`.

workflows/metatdenovo.nf
```diff
@@ -92,6 +92,7 @@ include { UNPIGZ as UNPIGZ_EUKULELE        } from '../modules/local/unpigz.nf'
 include { UNPIGZ as UNPIGZ_CONTIGS         } from '../modules/local/unpigz.nf'
 include { COLLECT_FEATURECOUNTS            } from '../modules/local/collect_featurecounts.nf'
 include { COLLECT_STATS                    } from '../modules/local/collect_stats.nf'
+include { SUM_EGGNOG                       } from '../modules/local/sum_eggnog.nf'
```
erikrikarddaniel: There isn't a subworkflow for eggnog? If there is, it would be better to hide this there I think.

erikrikarddaniel (commented):
Nice start. I have quite a few comments, particularly because you're mixing base R with Tidyverse.

workflows/metatdenovo.nf
danilodileo: I can do it!

modules/local/sum_eggnog.nf
danilodileo: yes I don't need it. It was just to select the element in the first table with the orfs in the second (so I am using only one column instead of all the 14 columns)

modules/local/sum_eggnog.nf
```diff
@@ -0,0 +1,61 @@
...
+    library(tidyverse)
+
+    TYPE_ORDER = c('sample', 'database', 'field', 'value')
+    # call the tables into variables
+    eggnog <- read_tsv("eggnogs.tsv.gz", show_col_types = FALSE ) %>%
+        as_tibble()
+
+    counts <- list.files(pattern = "*_counts.tsv.gz") %>%
+        map_df(~read_tsv(.,  show_col_types  = TRUE)) %>%
+            as_tibble()
```
erikrikarddaniel:
```suggestion
        as_tibble()
```

modules/local/sum_eggnog.nf
```diff
@@ -0,0 +1,61 @@
...
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    """
+    #!/usr/bin/env Rscript
+
+    library(data.table)
+    library(dtplyr)
```
erikrikarddaniel:
```suggestion
```

erikrikarddaniel (approved):
Two small things to look at, otherwise LGTM.

PR #117, opened 2023-03-20
schema.json updated

danilodileo:
I update the nextflow_schema.json file but I haven't find any guidelines about it on nf-core.


erikrikarddaniel (approved):
Let's hope this is good! :-)

PR #118, opened 2023-03-20
Fix kofamscan download output

danilodileo:
## PR checklist

I just changed the kofamscan_download module so now it will not stop (there was a mismatch between the output and the check in the code)

### DIFF
--- a/modules/local/kofamscan/download.nf
+++ b/modules/local/kofamscan/download.nf
@@ -27,7 +27,7 @@ process KOFAMSCAN_DOWNLOAD {
         gunzip -c ko_list.gz > ${kofam_dir}/ko_list
     fi
 
-    if [ ! -e ${kofam_dir}/koprofiles ]; then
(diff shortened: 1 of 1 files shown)


erikrikarddaniel (approved):
Good.

danilodileo:
@erikrikarddaniel  do you know anything about the last check? it is not solved after two days now so there is something wrong I think.

erikrikarddaniel:
> @erikrikarddaniel do you know anything about the last check? it is not solved after two days now so there is something wrong I think.

It's probably stuck. Happens every now and then. I think it's fine to bypass branch protection (haven't seen that before).

danilodileo:
> > @erikrikarddaniel do you know anything about the last check? it is not solved after two days now so there is something wrong I think.
> 
> It's probably stuck. Happens every now and then. I think it's fine to bypass branch protection (haven't seen that before).

I can't merge it, the button is locked because it is waiting the last check

PR #140, opened 2023-04-14
Make kofamscan use all cpus

erikrikarddaniel:
Add `--cpus` argument to the kofamscan module. *Not* tested :fearful:!


danilodileo (approved):
LGTM

PR #141, opened 2023-04-17
Merge tables bug

danilodileo:
## PR checklist

- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/metatdenovo/tree/master/.github/CONTRIBUTING.md)- [ ] If necessary, also make a PR on the nf-core/metatdenovo _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.


conf/modules.config
```diff
@@ -196,11 +196,11 @@ process {
         publishDir = [
             path: { "${params.outdir}/prodigal" },
             mode: 'copy',
-            pattern: "*.{faa,fna}"
+            pattern: "*.gz"
```
erikrikarddaniel, line 199: I'm hesitating here since you changed from only fasta files, but I suppose the `.gz` will include all output files. Fine, I think.

modules/local/format_prodigal.nf
```diff
@@ -20,9 +20,13 @@ process FORMAT_PRODIGAL {
...
     def args = task.ext.args   ?: ''
     prefix   = task.ext.prefix ?: "${meta.id}"
+    input    = gff =~ /\.gz$/ ? gff.name.take(gff.name.lastIndexOf('.')) : gff
+    gunzip   = gff =~ /\.gz$/ ? "gunzip -c ${gff} > ${input}" : ""
 
     """
-    sed 's/\\(\\(k[0-9]\\+_[0-9]\\+\\).*\\)ID=[0-9]\\+\\(_[0-9]\\+\\)/\\1ID=\\2\\3/g' $gff > ${prefix}_format.gff
+    $gunzip
+
+    sed 's/\\(\\(k[0-9]\\+_[0-9]\\+\\).*\\)ID=[0-9]\\+\\(_[0-9]\\+\\)/\\1ID=\\2\\3/g' $input > ${prefix}_format.gff
```
erikrikarddaniel: I think it would be better to build one unix pipeline and avoid creating an unzipped file. E.g.:

```
cat_input = gff =~ /\.gz$/ ? "gunzip -c ${gff} : "cat ${gff}"

"""
$cat_input | sed 's/\\(\\(k[0-9]\\+_[0-9]\\+\\).*\\)ID=[0-9]\\+\\(_[0-9]\\+\\)/\\1ID=\\2\\3/g' | gzip -c > ${prefix}_format.gff.gz
"""
```

Above, I _guessed_ that one wants the output gzipped, but that is maybe not the case, or? If not, just remove the `| gzip -c`. *Edit* Now I see that you do want the output zipped, so keep the full pipeline and remove the last two lines, see comment below too.

modules/local/format_prodigal.nf
```diff
@@ -20,9 +20,13 @@ process FORMAT_PRODIGAL {
...
     sed -i 's/\\(\\(NODE_[0-9]\\+\\).*\\)ID=[0-9]\\+\\(_[0-9]\\+\\)/\\1ID=\\2\\3/g' ${prefix}_format.gff
```
erikrikarddaniel: Why are you doing this here, better build a full pipeline above. (Remember to use `\\` to place commands on multiple lines.)

erikrikarddaniel (commented):
I'm a little puzzled regarding why you need the reformatting, but you know better than I... :-)

In any case, the code in that module can be simplified, see comments.

conf/modules.config
danilodileo: yes, it does. The new prodigal module has all .gz files

danilodileo:
> I'm a little puzzled regarding why you need the reformatting, but you know better than I... :-)
> 
> In any case, the code in that module can be simplified, see comments.

The main fasta file output had different name then the gff ids. So the functional tables will not match with the counts tables as they have different reference (one has the gff while the other has the protein fasta file).

erikrikarddaniel (approved):
:+1:

Issue #151, opened 2023-06-12
`merge_tables` and `collect_stats` fails if sample names are just digits


emnilsson:
### Description of the bug

As the description says, `merge_tables` fails when the sample names are just digits, e.g. "101". `union()` complains about `! x and y are not compatible.
✖ Incompatible types for column sample: character vs double.`. `collect_stats` fails due to similar reasons. Could we either make it possible to have samples that are named with just digits, or throw an error at the sample-check?

danilodileo:
Solved with PR #152

PR #152, opened 2023-06-27
sample name can be a number or a character and it will not crush at c…

danilodileo:
…ollect stats step

## PR checklist

- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/metatdenovo/tree/master/.github/CONTRIBUTING.md)- [ ] If necessary, also make a PR on the nf-core/metatdenovo _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.

This PR deals with #151

### DIFF
--- a/conf/modules.config
+++ b/conf/modules.config
@@ -406,7 +406,7 @@ process {
         publishDir = [
             path: { "${params.outdir}/summary_tables" },
             mode: 'copy',
-            pattern: '*.tsv'
+            pattern: '*.tsv.gz'
         ]
     }
 
--- a/conf/modules.config
+++ b/conf/modules.config
@@ -378,9 +378,7 @@ process {
 
     withName: SUM_EGGNOG {
         publishDir = [
-        path: { "${params.outdir}/summary_tables" },
-        mode: 'copy',
-        pattern: "eggnog_summary.tsv"
+            enabled: false
         ]
     }
 
@@ -398,7 +396,6 @@ process {
             path: { "${params.outdir}/eukulele" },
             mode: 'copy'
         ]
-        memory = 16.GB
         ext.args = [
             params.orf_caller == 'transdecoder' ? "--protein_extension .pep" : '',
             params.eukulele_method ? "-m ${params.eukulele_method}" : ''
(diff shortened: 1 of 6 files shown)


conf/modules.config
```diff
@@ -378,9 +378,7 @@ process {
 
     withName: SUM_EGGNOG {
         publishDir = [
-        path: { "${params.outdir}/summary_tables" },
-        mode: 'copy',
-        pattern: "eggnog_summary.tsv"
+            enabled: false
```
erikrikarddaniel, line 381: Are these intermediary files that should not be saved? (Same below.)

modules/local/merge_summary_tables.nf
```diff
@@ -31,9 +31,9 @@ process MERGE_TABLES {
     library(tidyr)
     library(stringr)
 
-    summary <- data.frame(sample = character(), database = character(),field = character(),  value = numeric(), stringsAsFactors = FALSE) %>%
+    summary <- data.frame(sample = character(), database = character(), field = character(), value = numeric(), stringsAsFactors = FALSE) %>%
         union(list.files(pattern = "*.tsv.gz") %>%
-        map_df(~read_tsv(.,  show_col_types  = TRUE))) %>%
+            map_df(~ read_tsv(., show_col_types = TRUE) %>%  mutate(sample = as.character(sample)))) %>%
```
erikrikarddaniel: Indentation seems off, should be the same as `union()` above and `pivot_wider()` below.

erikrikarddaniel (approved):
A small indentation thing, plus I'm not sure what's going on in `modules.conf` with the saving. As I suspect you know better than I, I'm approving anyway.

conf/modules.config
danilodileo: yes, we decided to not have them in the folder as they are with all other info in overall_stats.tsv.gz

modules/local/merge_summary_tables.nf
danilodileo: the two map_df are inside union function, so both should be indent.

PR #167, opened 2023-08-24
Issue 162 add assembler qc report

tfalkarkea:
## PR checklist

- [x] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/metatdenovo/tree/master/.github/CONTRIBUTING.md)
- [x] `README.md` is updated (including new tool citations and authors/contributors).

## Comments
Added a module for transrate, which is a simple way to create some quality control metrics for an assembly file. The CSV transrate generates is then included in the final MultiQC, which has some minor formatting adjustments included in `assets/multiqc_config.yml`.

Further changes may be useful to this MultiQC formatting.


CITATIONS.md
```diff
@@ -73,8 +73,12 @@
 - [EUKulele](https://github.com/AlexanderLabWHOI/EUKulele)
 
 - [CAT](https://github.com/dutilh/CAT)
+
```
erikrikarddaniel, line 76: Should this empty line be here? You don't seem to have added one after transrate below.

modules/local/transrate.nf
```diff
@@ -0,0 +1,42 @@
...
+
+    // transrate flashes a warning about a ruby gem being out of date, so call the version before it is being piped into the yaml
+    """
+    gzip -d -c $assembly > assembly_unzipped.fa
+
+    transrate \\
+        --threads $task.cpus \\
+        --assembly assembly_unzipped.fa \\
+        --output ${prefix}_transrate \\
+        $args
```
erikrikarddaniel: A few thoughts:
1. nf-core modules should take gzipped files, so from that point of view unzipping in the module is fine.
2. For other purposes, one could call the `UNPIGZ` module (local to this workflow) or the nf-core `GUNZIP` module before instead. I know that they preferred that in the mag pipeline because more than one module wanted unzipped input and by calling `UNPIGZ` they were left with only one unzipped file instead of one in each work directory where it was needed. _Edit:_ Now I see that we actually call `UNPIGZ`. You should use the output channel from that as input to this and remove the `gunzip`.
3. You can't trick `transrate` to read a compressed file with `--assembly <(gunzip -c $assembly)`?

workflows/metatdenovo.nf
```diff
@@ -317,6 +318,13 @@ workflow METATDENOVO {
         ch_versions = ch_versions.mix(MEGAHIT_INTERLEAVED.out.versions)
     }
 
+    //
+    // MODULE: Use TransRate to judge assembly quality
+    //
+    TRANSRATE(ch_assembly_contigs)
```
erikrikarddaniel:
```suggestion
    TRANSRATE(UNPIGZ_CONTIGS.out.unzipped)
```
See comment above.

erikrikarddaniel (commented):
I think the module would be a good addition to the nf-core set of modules. A little more work, as you're required to provide tests, but usually worth it.

I have a question regarding whether the `TRANSRATE` module should unzip or not, see comment in the code. If the tool can be made to read gzipped files, that would be best, if not, let it take an unzipped file and feed it with the output from the `UNPIGZ_CONTIGS` module.

CITATIONS.md
tfalkarkea: Looks like most of them have an empty line between link and full citation. I did neglect one for transrate, but the markdown omits them either way.

erikrikarddaniel: OK.

modules/local/transrate.nf
tfalkarkea: I am a little confused here, you say the transrate module is good to take in gzipped files, but not in this case of the meta-t pipeline, since we're passing in UNPIGZ?

tfalkarkea: Also, if we're using `UNPIGZ_CONTIGS.out.unzipped`, we should use the transrate module in the `CAT` block on line 455 yes?

tfalkarkea: I have seen some other modules test for gzipped-ness and proceed accordingly, that might be the best general purpose move here.

erikrikarddaniel: Sorry, I meant the first comment as a discussion start so it got a bit confusing. On top of that, I had missed that the `UNPIGZ_CONTIGS` was inside the `CAT` part, meaning it won't be available to `TRANSRATE`.

There are a couple of routes we could take: 1) Keep your code as it originally was, 2) move the call to `UNPIGZ_CONTIGS` to outside the `CAT` context and change `TRANSRATE` so it takes unzipped files. (I thought of another alternative, but can't remember now...)

I'm rather fond of number 2. `TRANSRATE` will always be called so unzipping will be needed at least once, so we're not unzipping unnecessarily. If the user decides to use `CAT`, we're saving one unzipping operation. There might be other modules that could be called with the output from `UNPIGZ_CONTIGS` (that take either zipped or unzipped). We could look for such cases later to save on unzipping.

If you decide to contribute the module to nf-core, you might want to rewrite it to take either zipped or unzipped. We can then keep the call as it is, but if we decide to let `CAT` go (which we might well do) we can drop `UNPIGZ_CONTIGS`. (In this case, I'm fond of the `<(gunzip -c file.gz)` syntax as it doesn't leave traces.)

tfalkarkea: I agree, the second option is the right way to go! I implemented and it seems to run smoothly, take a look and see if I changed what you had in mind.

modules/local/transrate.nf
```diff
@@ -0,0 +1,41 @@
...
+    path "versions.yml"                          , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    // transrate flashes a warning about a ruby gem being out of date, so call the version before it is being piped into the yaml
```
erikrikarddaniel: I would put this comment before the version call (with `#` in front of course).

erikrikarddaniel (approved):
Looks fine. Just a tiny comment.

modules/local/transrate.nf
tfalkarkea: Sure! Not a problem. Thanks for the input!

danilodileo (approved):
LGTM

PR #178, opened 2023-08-30
Fix filenames

erikrikarddaniel:
## PR checklist

The point of this was to make sure that output files in `results/summary_tables` all had the assembler and orf caller as prefixes. However, I also did some cleaning up and refactoring, so the PR is not as clean as it should be.


tfalkarkea (approved):
Looks good, mostly cleaning up some code but glad to see some more specifics in inputs and outputs naming.

erikrikarddaniel:
Thanks!

PR #196, opened 2023-10-19
Fix eggnog download no copy ode

danilodileo:
## PR checklist

I deleted the cp command. Now the database is only stored in dbpath

### DIFF
--- a/subworkflows/local/eggnog.nf
+++ b/subworkflows/local/eggnog.nf
@@ -20,7 +20,7 @@ workflow EGGNOG {
 
         if ( ! eggnogfile.exists() ) {
             EGGNOG_DOWNLOAD(Channel.fromPath(eggnog_dbpath))
-            ch_dbpath = EGGNOG_DOWNLOAD.out.db
+            ch_dbpath = EGGNOG_DOWNLOAD.out.db_ready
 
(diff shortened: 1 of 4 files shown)


subworkflows/local/eggnog.nf
```diff
@@ -20,7 +20,7 @@ workflow EGGNOG {
 
         if ( ! eggnogfile.exists() ) {
             EGGNOG_DOWNLOAD(Channel.fromPath(eggnog_dbpath))
-            ch_dbpath = EGGNOG_DOWNLOAD.out.db
+            ch_dbpath = EGGNOG_DOWNLOAD.out.db_ready
```
erikrikarddaniel: So, this channel is not emitted right away but only when the download program finishes? In a way, the old name of the corresponding channel, i.e. `db`, would have been more conservative and this change wouldn't be required.

erikrikarddaniel (approved):
Fine, but with a small comment. Do I understand this correctly as meaning that the database directory channel _is not_ emitted until the download finishes and you had misunderstood? (I just want to make sure I understand Nextflow correctly.)

subworkflows/local/eggnog.nf
danilodileo: Yes. we can keep the old name (already updated)

PR #428, opened 2026-03-16
Template update 3.5.2

danilodileo:
## PR checklist

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/metatdenovo/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, also make a PR on the nf-core/metatdenovo _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [ ] Make sure your code lints (`nf-core pipelines lint`).
- [ ] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [ ] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [ ] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).

I updated the template following the guidelines. Modules need to be updated to but I will do it in the next PR after this one.

PR #438, opened 2026-06-11
fix(eukulele): support mags mode by using wildcard instead of hardcoded mets_full

TeresaRubio:
When running EUKulele with -m mags, the output is generated in 
     mags_full/diamond/ instead of mets_full/diamond/, causing a gzip error.
     
     Fix: replace hardcoded mets_full with *_full wildcard to support both
     mets and mags modes.


erikrikarddaniel (changes requested):
Looks good. Please add an entry to `CHANGELOG.md`. (Don't forget your github tag in parentheses after the message (see e.g. #428) -- it's nice to keep track of who did what.)

Issue #456, opened 2026-08-11
Add more gene caller info to MultiQC report


erikrikarddaniel:
### Description of feature

I just got a suggestion to "add stats of prodigal called proteins on the multiqc report of metatdenovo". Sounds like a very good idea, that we could also implement for the other two gene callers.

erikrikarddaniel:
Done in #458

Issue #411, opened 2025-09-03
Allow multiple assemblies


erikrikarddaniel:
### Description of feature

The primary use case for this is when it's difficult/impossible to make a single assembly because of resource constraints.

Assemblies could be per sample or per groups of samples in the style of nf-core/mag.

After assembly: cluster (Vsearch, mmseqs?) feature (e.g. ORF) sequences. Should this be done prior to quantification or after? The former would suggest we need to map to the feature sequences rather than the contigs, the latter would leave a translation file between feature identifiers in the different assemblies.

An alternative is perhaps to make a meta-assembly. The tools I find seem more tailored to single genomes than meta-omics.

erikrikarddaniel:
This issue's own open question -- "Should [feature/ORF clustering] be done prior to quantification or after?" -- came up again from a different angle in #460 (consolidating ORF/protein calls across multiple gene callers, e.g. Prokka/Prodigal for prokaryotic content vs. TransDecoder/MetaEuk for eukaryotic content in a mixed-community project). The reasoning there transfers directly here:

Cluster first, at the protein level, before annotation and quantification. Annotate/quantify once per cluster representative rather than once per redundant call, then trace read counts and annotation back to every member of the cluster via a cluster-ID mapping. This avoids both wasted redundant compute (re-running eggnog/kofamscan/dbcan/diamond on near-duplicate sequences) and the messier problem of two different tools/scores disagreeing for what's supposedly the same gene.

This issue's own "path 3" (cluster features, map reads to the cluster representative) is structurally the same solution proposed for #460 -- if that gets built, it's likely directly reusable here rather than needing a second, independently-designed clustering mechanism.

One reason to lean toward clustering at the *protein* level specifically (not contigs, "path 1" above): in the per-sample/per-group assembly scenario this issue describes, strain-level nucleotide variation between samples would often still translate to the same or near-identical protein, so protein-level clustering tolerates that "for free" -- the same property that makes it work across #460's genomic-vs-transcript sequence differences.

One thing that doesn't transfer directly: this issue's redundancy is per-sample, potentially at much larger scale than #460's per-caller redundancy (at most a handful of overlapping call sets there). Whatever clustering tool/threshold gets chosen needs to perform well at this issue's scale too, not just be tuned for #460's smaller case.

Moved both this issue and #460 to the 2.0 milestone -- this clustering mechanism is more involved than what fits in 1.5.0.

erikrikarddaniel:
[...]

PR #458, opened 2026-08-12
Fix Prokka MultiQC stats collision, add Prodigal/TransDecoder MultiQC stats

erikrikarddaniel:
## PR checklist

## Description

Addresses #456 in full (both halves).

**Prokka fix**: `PROKKA_SUBSETS` runs one Prokka invocation per contig batch (`splitFasta`), so any assembly larger than `params.prokka_batchsize` produces multiple `.txt` summaries. Prokka is never given `--genus`/`--species`/`--strain`, so every batch's summary carries the identical literal `organism: Genus species strain ` line, and MultiQC's native `prokka` module derives its sample name from that line -- so on any real multi-batch assembly, all batches collided under a single `strain` sample in the report, with only the last-parsed batch's stats surviving. Fixed by summing the numeric fields across all batches into one per-assembly summary with a real sample name. Verified by forcing 5 batches (`--prokka_batchsize 500000`) -- all sums match the per-batch totals exactly, including a batch missing the `rRNA:` line entirely (handled as 0, no crash).

**Prodigal/TransDecoder stats**: neither has a MultiQC-native module or an existing summary file, so added a small custom-content CSV (`n_orfs`, `total_aa`, `mean_aa_length`) built directly from each tool's own amino-acid FASTA, via the same `collectFile` pattern already used for TransRate's assembly stats. Verified against real `-profile test` (prodigal) and `-profile test_megahit_transdecoder` runs.

No nf-test coverage added for either fix (matches existing precedent -- e.g. eggnog/eukulele also have no dedicated pipeline-level nf-test); verification was done via direct pipeline runs and inspecting the real `multiqc_data/` output.

Generated by Claude, reviewed by @erikrikarddaniel before opening.


subworkflows/local/prokka/subsets/main.nf
```diff
@@ -24,7 +24,40 @@ workflow PROKKA_SUBSETS {
...
-    ch_log  = PROKKA.out.txt.map { _meta, log -> log }.collect()
+    // PROKKA runs once per contig batch (see splitFasta above), so on any assembly larger
+    // than params.prokka_batchsize there are multiple *.txt summaries, each with the same
+    // literal "organism: Genus species strain " line (no --genus/--species/--strain is
+    // passed to PROKKA). MultiQC's prokka module derives the sample name from that line, so
+    // feeding the raw per-batch files in directly makes every batch collide under the same
+    // "strain" sample name. Sum the numeric fields across all batches into a single
+    // per-assembly summary instead, with a real sample name.
+    ch_log = contigs
+        .map { meta, _contigs -> meta.id }
```
tracelail, line 35: Does this subworkflow ever process more than one assembly in a single pipeline run (e.g. multiple samples/assemblies flowing through in parallel)? The new summing logic pairs each assembly ID with the entire collected list of txt files via .combine(...collect()...). If only one assembly is ever in flight at a time, this is a non-issue but if multiple assemblies can be processed together, I just wanted to confirm the code filters each assembly down to only its own batch files before summing, rather than accidentally summing another assembly's batch totals into this one's summary.

workflows/metatdenovo.nf
```diff
@@ -413,6 +413,29 @@ workflow METATDENOVO {
...
+                reader.eachLine { line ->
+                    if (line.startsWith('>')) {
+                        n_orfs = n_orfs + 1
+                    } else {
+                        total_aa = total_aa + line.trim().length()
+                    }
+                }
+                reader.close()
+                def mean_aa = n_orfs > 0 ? (total_aa / n_orfs) : 0
+                def content = "Sample,n_orfs,total_aa,mean_aa_length\n${meta.id},${n_orfs},${total_aa},${String.format(Locale.ROOT, '%.1f', mean_aa)}\n"
```
tracelail, line 435: Did you try testing with 0 n_orfs? I checked the `tests/spades_prodigal.nf.test` and there's no test for the fall back you implemented of: `def mean_aa = n_orfs > 0 ? (total_aa / n_orfs) : 0`

tracelail (commented):
Logic looks good to me, Two, likely minor, items to confirm.

subworkflows/local/prokka/subsets/main.nf
erikrikarddaniel: Good question, and no — today this subworkflow is only ever invoked with a single assembly per pipeline run. `ch_assembly_contigs` (built upstream in `workflows/metatdenovo.nf`, from MEGAHIT/SPAdes/user-assembly) is always a single-item channel across every code path — this pipeline combines all samples' reads into one co-assembly, it doesn't currently support per-sample or multiple independent assemblies in a single run. So `contigs.map{meta,_contigs -> meta.id}` here is always a single value, and the `.combine()` pairing is always 1:1 (one assembly ID × its own batch files) — no risk of summing another assembly's totals into this one.

Worth noting for the future: multi-assembly support is tracked as a separate feature (#411, targeted for 2.0) — if/when that lands, this summing logic (and a few other places that assume a single global assembly) would need revisiting. Not a concern for this PR though.

(reply drafted with some help from Claude Code)

workflows/metatdenovo.nf
erikrikarddaniel: Good catch — no, I didn't add a dedicated test for the 0-ORF fallback, and you're right the existing tests never exercise it (every fixture assembly produces plenty of real ORFs). I checked what happens without the guard: `total_aa / n_orfs` with `n_orfs=0` does throw (confirmed: `ArithmeticException: Division undefined`), so the ternary is a real, necessary guard, not defensive-for-no-reason.

Getting a genuine end-to-end test for this would need a pathological input that makes Prodigal/TransDecoder call zero ORFs on an otherwise-successful assembly, which none of the current test fixtures naturally produce — open to suggestions if you know a good way to force that, otherwise I'd lean toward leaving this as a verified-but-not-independently-tested defensive guard, similar to other small edge-case branches elsewhere in the codebase. Let me know if you'd rather see it covered properly before merge.

(reply drafted with some help from Claude Code)

subworkflows/local/prokka/subsets/main.nf
tracelail: That makes sense. This sounds like a non-issue and good to know about #411.

workflows/metatdenovo.nf
tracelail: I don't have a good way to force a pathological zero-ORF case myself (I'm outside my depth on the Prodigal/TransDecoder internals), so I'll defer to your judgement on whether "verified-but-not-independently-tested" is fine here.

tracelail (commented):
I agree that the single assembly point is a non-issue but I'll defer to your judgement about the 0-orf.

erikrikarddaniel:
@tracelail : I think you forgot to approve, thanks!

tracelail (approved):

tracelail:
> @tracelail : I think you forgot to approve, thanks!

I did! Should be approved now.

Issue #462, opened 2026-08-15
Support running multiple ORF callers in a single pipeline execution


erikrikarddaniel:
## Summary

Today `--orf_caller` is a single global choice (`prodigal`/`prokka`/`transdecoder`), applied to the whole assembly for the entire pipeline run. Consolidating ORF calls across callers (#460) requires running more than one caller in the same execution first -- this issue is that prerequisite.

## Scope

- Generalize `orf_caller` handling so more than one caller can be active in a single run (e.g. prokka + transdecoder, or prokka + metaeuk once #459 lands).
- Each active caller keeps running exactly as today, independently, producing its own GFF/protein FASTA/`summary_tables` output (the existing per-caller tables) -- no consolidation logic in this issue, just parallel execution of multiple callers.
- Existing single-caller behavior/output must be unchanged when only one caller is selected.

## Out of scope (tracked separately)

- Locus-level (same-contig, cross-caller) consolidation
- Cluster-level (cross-contig) consolidation: #460

## Related

#459 (MetaEuk gene caller) is not a hard dependency for this issue -- Prokka+TransDecoder is enough to build and test the multi-caller execution path -- but is needed for the actual motivating mixed prokaryote/eukaryote use case that #460 is about.

erikrikarddaniel:
Done in #466

Issue #460, opened 2026-08-12
Consolidate ORF/protein calls across multiple gene callers in mixed prokaryote/eukaryote projects


erikrikarddaniel:
## Summary

Follow-on from #459 (MetaEuk as an optional eukaryote-aware gene caller). Splitting this out since it's a genuinely separate, harder question that shouldn't block MetaEuk's addition there.

Today, `--orf_caller` is a single global choice for the whole pipeline run (prodigal/prokka/transdecoder), so the pipeline never has to reconcile ORFs from different callers -- there's only ever one set. That stops being true once a mixed prokaryotic+eukaryotic community project wants prokaryotic content called with Prokka/Prodigal *and* eukaryotic content called with TransDecoder and/or MetaEuk in the same analysis. Two related problems come out of that:

1. **Linking a genomic (splice-aware) gene call to a transcript-derived call for the same underlying gene.** MetaEuk reconstructs a spliced CDS/protein directly from genomic-style contigs by chaining exons across introns using homology to a reference protein database. A TransDecoder call on a corresponding transcript for the same gene is intron-free by construction (RNA is already spliced). So a genomic MetaEuk call and a transcript TransDecoder call for the same real gene should converge to a highly similar or identical *protein sequence*, even though they come from entirely different contigs with no coordinate relationship to each other. That makes this a protein-similarity clustering problem, not a genome-alignment/coordinate-linking problem -- the intron-awareness only needs to happen once, inside MetaEuk itself.

2. **Consolidating prokaryotic (Prokka/Prodigal) and eukaryotic (TransDecoder/MetaEuk) calls into one non-redundant ORF/protein set.** Same underlying mechanism as (1): cluster all called proteins across every active caller by sequence similarity (e.g. MMseqs2 `cluster`/`linclust` -- already a natural fit since MetaEuk itself is MMseqs2-based); whatever falls in one cluster is treated as the same gene/protein regardless of which tool or domain-of-life path produced it.

Taxonomy is *not* proposed as the primary reconciliation key here, deliberately: the ORFs that motivated #459 in the first place are exactly the ones with no taxonomic hit at all (see #459's motivation section). If taxonomy were the primary grouping mechanism, it would specifically fail on the population of genes we most want to capture correctly. Taxonomy could still be a useful secondary sanity check (e.g. flag a cluster as suspicious if it spans a confidently-prokaryotic and a confidently-eukaryotic annotation), but sequence identity should carry the real weight.

[...]

PR #466, opened 2026-08-19
Support running multiple ORF callers in a single pipeline execution

erikrikarddaniel:
## PR checklist

## Description

This is part of #460 which will give us much better support for mixed communities, particularly with large proportions of eukaryotes.

Addresses #462: `--orf_caller` now accepts a comma-separated list (e.g. `--orf_caller prokka,transdecoder`) to run more than one ORF caller in the same execution. Each active caller runs fully independently and produces its own complete `summary_tables/<assembly>.<caller>.*` output, exactly as if it had been the only caller selected -- no cross-caller consolidation yet (tracked separately in #460/#463).

### Why this touched so much

Nearly everything downstream of ORF calling (`FEATURECOUNTS_CDS`, `COLLECT_FEATURECOUNTS`, `MERGE_TABLES`, `COLLECT_STATS`, the `EGGNOG`/`KOFAMSCAN`/`DBCAN`/`EUKULELE`/diamond-taxonomy annotation subworkflows, `HMMCLASSIFY`) was written assuming exactly one caller per run -- either reassigning a single shared channel variable, or doing a flat, un-keyed `.collect()`/`.combine()`/`.join()` across everything. Making N callers coexist without cross-contaminating each other's results meant threading caller identity through as an explicit `caller` field on each item's meta map, and regrouping/rejoining most of that downstream code per caller instead of per run.

Three real, verified bugs came out of building and testing this (all fixed, all with regression coverage):

- **`UNPIGZ_GFF` invoked twice in one run.** Prokka, Prodigal, and MetaEuk each called this shared, unaliased process from their own `if` block -- fine when only one could ever be true, fatal (`Process 'UNPIGZ_GFF' has been already used`) once two of them run together. Fixed by unzipping every active caller's GFF through a single call. Found via an extra code review pass on this branch, not by the original tests -- the one multi-caller test at the time paired prokka+transdecoder, and transdecoder never touches this process. Now covered directly: `tests/megahit_prokka_prodigal_transdecoder.nf.test` runs prokka+prodigal+transdecoder together.
- **`MERGE_TABLES` crashing on zero input tables.** An earlier version of this change force-invoked it once per caller even when that caller had no annotation-subworkflow output, which its R script isn't written to handle. Fixed to only invoke it for callers that actually have >=1 table, matching the pre-existing single-caller behavior of not invoking it at all when every annotation subworkflow is skipped.
- **`FEATURECOUNTS_CDS` losing `single_end`.** Its caller-suffix meta rebuild replaced the sample's meta wholesale instead of merging into it, dropping `single_end`/`strandedness` and breaking the `-p` flag selection for single-end samples.

### Other consequences worth flagging explicitly

- `featurecounts/*.featureCounts.tsv` output filenames are now always caller-suffixed (e.g. `SAMPLE1.prokka.featureCounts.tsv`), including in single-caller runs -- the one place a single-caller filename changes, needed to stop multiple simultaneous callers overwriting each other's published per-sample counts. All 8 pre-existing single-caller snapshot tests were re-verified and updated for exactly this diff, nothing else.
- Four `conf/modules.config` blocks (`KOFAMSCAN_SCAN`, `KOFAMSCAN_FORMAT`, `DBCAN_FORMAT`, `FORMAT_EUKULELE_TAX`) hardcoded `params.orf_caller` directly into output filenames, bypassing `meta.id` -- with a comma-joined multi-value `orf_caller` this produced literal commas in filenames and cross-caller collisions. Switched to `meta.id`, which already carries the correct per-caller value.

### Testing

- New `tests/megahit_prokka_prodigal_transdecoder.nf.test`: prokka+prodigal+transdecoder on the same input samplesheet/assembler the existing solo `tests/megahit_prokka.nf.test`, `tests/default.nf.test` (prodigal), and `tests/megahit_transdecoder.nf.test` already use, so its per-caller assertions are directly comparable to those tests' own recorded values as independent confirmation that no caller's output is contaminated by another's. Also asserts zero ORF-ID overlap between every pair of callers, and that every active caller's `featurecounts`/`hmmrank` output exists independently. Verified stable across two independent runs.
- All 8 pre-existing single-caller pipeline tests re-run and confirmed green (only the expected `featurecounts` filename diff in their snapshots).
- `prek run -a`, `nf-core pipelines lint`, and `nextflow lint .` (both the declared minimum Nextflow version and latest) all clean.


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Added`
 
+- [#466](https://github.com/nf-core/metatdenovo/pull/466) - Support running multiple ORF callers in a single execution, e.g. `--orf_caller prokka,transdecoder` (each caller still produces its own independent output, no cross-caller consolidation yet), addresses [#462](https://github.com/nf-core/metatdenovo/issues/462) (@erikrikarddaniel)
```
danilodileo, line 10: Is cross-caller the final aim?

danilodileo (approved):
This is a nice improvement that will benefit many of us. thanks!

Issue #463, opened 2026-08-15
Consolidate same-contig ORF calls from different callers into locus-level counts


erikrikarddaniel:
## Summary

When multiple ORF callers run on the same assembly (#462), they can independently annotate the same or overlapping genomic coordinates on the same contig as separate features -- e.g. Prokka's CDS and TransDecoder's ORF for the same real gene. Counting reads against each caller's GFF independently and summing the results afterwards double-counts every read supporting that locus: each caller's `featureCounts` run is blind to the other callers' claim on the same interval, so the same aligned read gets counted once per caller rather than once total.

## Background

From discussion on #460: this is a distinct problem from cross-contig consolidation (different assemblies/contigs entirely, e.g. a genomic vs. a transcript assembly of the same gene). There, per-read alignment exclusivity across contigs already prevents double counting when summing counts after the fact -- a read can only align to one contig. Same-contig, cross-caller overlap is different: the coordinate-level merge has to happen *before* counting, not after, because the ambiguity is at the feature-assignment level, not the alignment level.

## Scope

- Given all active callers' GFFs for the same assembly, detect overlapping/identical-coordinate features across callers, per contig (e.g. via interval merging, `bedtools merge`-style).
- Merge overlapping same-locus features from different callers into one interval; count reads once per locus, not once per contributing caller.
- Produce a new locus-consolidated counts table (naming TBD, e.g. `<assembly>.locus_consolidated.counts.tsv.gz`), with provenance columns: which caller(s), and how many independent calls, support each locus.
- When only one caller is active, this table should be identical in content to that caller's own existing per-caller table -- graceful degradation, not a separate code path for the single-caller case.

## Out of scope (tracked separately)

- Cross-contig/cross-assembly consolidation (protein-sequence clustering): #460

## Depends on

- #462 (multiple ORF callers in a single run)

erikrikarddaniel:
Done in #467

PR #467, opened 2026-08-20
Consolidate overlapping multi-caller ORF loci before counting

erikrikarddaniel:
## PR checklist

## Description

Addresses #463.

When more than one `--orf_caller` is active (see #462/#466), each caller currently produces its own completely independent output table.
If two active callers call overlapping or identical-coordinate CDS features on the same contig for what's really the same gene, summing their tables double-counts every read supporting that locus.

This PR adds a new, additional output, `summary_tables/<assembly>.locus_consolidate.counts.tsv.gz`, where overlapping same-contig, same-strand calls from different callers are merged into one locus **before** counting.
Each active caller's own per-caller table and counting still happens exactly as before -- this table is produced alongside them, not instead of them.

### Design

- `FORMAT_GFF2BED` converts each active caller's CDS GFF to BED6, tagging the name column `<caller>:<original ID>`.
- All callers' BED output is concatenated, sorted, and merged with `bedtools merge -s` (strand-aware, so same-position opposite-strand calls are correctly kept separate).
- `FORMAT_LOCUS_CONSOLIDATE` turns the merged intervals back into a GFF, applying an ID-inheritance rule: a locus with exactly one distinct contributor keeps that contributor's own original ID verbatim; a locus with several distinct contributors gets a fresh coordinate-derived ID. This is what makes the feature degrade gracefully -- **with a single caller active, every merged locus has exactly one contributor, so the consolidated table is identical in content to that caller's own table** (verified in `tests/metaeuk.nf.test`).
- The consolidated GFF rides through the pipeline's existing, unmodified `FEATURECOUNTS_CDS` cross-product as just another "caller," so counting logic itself needed no changes.
- `COLLECT_LOCUS_CONSOLIDATE` (a near-clone of the existing `COLLECT_FEATURECOUNTS`) joins in a provenance table (`callers`, `n_calls` columns) recording which caller(s) contributed to each locus and how many independent calls were merged.
- `COLLECT_FEATURECOUNTS` and `FEATURECOUNTS_CDS` themselves are completely untouched -- zero risk to any existing per-caller output.

### Testing

- New `tests/metaeuk_transdecoder.nf.test`: on real overlapping MetaEuk+TransDecoder calls on the RPL28 test data, asserts the consolidated table reports the true (not double-counted) read total, correct `callers`/`n_calls` provenance on the merged locus, and that a same-position-opposite-strand pair of calls correctly stays split into two loci.
- Extended `tests/metaeuk.nf.test` to assert the single-caller degeneracy property explicitly.
- All other existing pipeline tests' snapshots updated, since this step now runs (as a no-op merge) for every caller combination, not just multi-caller runs.

### Post-implementation review

Ran a focused self-review pass after the initial implementation (independent of the main implementation work), which surfaced two issues fixed here:

- `COLLECT_LOCUS_CONSOLIDATE`'s original wiring used an unkeyed `.combine()` between the featureCounts and provenance channels -- safe today only because both are singleton channels (one assembly per run), but a real hazard if the pipeline is ever extended to multiple simultaneous assemblies. Switched to a `.join()` keyed on `meta.id`.
- Found and fixed a pre-existing, unrelated dead validation check (`--orf_caller`/`--user_orfs_gff` mutual-exclusivity, tautological condition that could never fire) while tracing how `ch_gff` is populated.

Not in scope (per #463 itself): cross-contig consolidation (tracked separately, #460).

### DIFF
--- a/modules/local/format/gff2bed/main.nf
+++ b/modules/local/format/gff2bed/main.nf
@@ -1,0 +1,48 @@
+process FORMAT_GFF2BED {
+    tag "$meta.id"
+    label 'process_low'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/gzip:1.11':
+        'biocontainers/gzip:1.11' }"
+
+    input:
+    tuple val(meta), path(gff)
+
+    output:
+    tuple val(meta), path("${prefix}.bed"), emit: bed
+    tuple val("${task.process}"), val('gzip'), eval('gzip --version  2>&1 | grep "^gzip" | sed "s/^gzip \\([0-9.]\\+\\).*/\\1/"'), emit: versions_gzip, topic: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    prefix    = task.ext.prefix ?: "${meta.id}"
+    cat_input = gff =~ /\.gz$/ ? "gunzip -c ${gff}" : "cat ${gff}"
+
+    // BED is 0-based half-open, GFF is 1-based inclusive: start shifts by one, end is unchanged.
+    // The "(^|;)ID=" anchor (rather than a bare "ID=") avoids matching inside MetaEuk's
+    // Target_ID=/TCS_ID= attributes, which also contain the substring "ID=".
+    // The "cds." prefix strip mirrors COLLECT_FEATURECOUNTS's existing TransDecoder-ID normalization
+    // (modules/local/collect/featurecounts/main.nf), so a locus this caller is the sole contributor to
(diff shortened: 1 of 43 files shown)


modules/local/format/gff2bed/main.nf
```diff
@@ -0,0 +1,48 @@
+process FORMAT_GFF2BED {
```
danilodileo, line 1: I believe (but I am probably wrong) there is a bedtools nf-core modules. couldn't it that one be adapted for the purpose of this PR?

danilodileo (approved):
LGTM

modules/local/format/gff2bed/main.nf
erikrikarddaniel: This is just reformatting. The nf-core modules for bedtools itself are used.

Issue #464, opened 2026-08-15
Expose --bbmap_ambiguous and --featurecounts_fraction (port from nf-core/magmap)


erikrikarddaniel:
## Summary

nf-core/magmap already exposes BBMap's ambiguous-read handling mode and a matching featureCounts fractional-counting toggle:

- `--bbmap_ambiguous` (enum: `best`/`all`/`random`/`toss`, default `best`) -> `ambig=` in `BBMAP_ALIGN`'s `ext.args`
- `--featurecounts_fraction` (boolean) -> pairs with it for the featureCounts step, magmap's own docs note it "works best coupled with `--bbmap_ambiguous`"

metatdenovo currently has neither -- `BBMAP_ALIGN` and `FEATURECOUNTS_CDS` both run with tool defaults (single-best-hit alignment, primary-alignment-only counting), with no way for a user to change this.

## Why this matters for consolidation (#460, #463)

The locus- and cluster-level consolidation design assumes each read is exclusively assigned to one ORF at counting time, so that summing per locus/cluster after the fact doesn't double- or under-count. That assumption holds for BBMap's default (`best`) and for `random`/`toss`, but breaks if a user sets `ambig=all` without also enabling fractional counting -- a single read could then be independently assigned to multiple contigs' ORFs, and naive summing across a cluster would multi-count it.

## Scope

- Add `--bbmap_ambiguous` (same enum/default as magmap) and `--featurecounts_fraction`, matching magmap's existing implementation/docs where reasonable.
- Document the compatibility requirement explicitly in `docs/usage.md`: `best`/`random`/`toss` are safe for consolidation as-is; `all` requires `--featurecounts_fraction` to stay safe, and this should be called out wherever the consolidated output tables are documented too.

## Depends on

- None -- self-contained, can ship independently and doesn't need to wait on #462/#463/#460.

erikrikarddaniel:
Shipped in #468 (merged 2026-08-25) — `--bbmap_ambiguous` and `--featurecounts_fraction` are both in place. Closing; the PR just didn't use a closing keyword.

PR #468, opened 2026-08-21
Expose --bbmap_ambiguous and --featurecounts_fraction

erikrikarddaniel:
## PR checklist

## Description

Addresses #464.

Ports `--bbmap_ambiguous` and `--featurecounts_fraction` from nf-core/magmap, which already exposes both. Neither existed in metatdenovo before this PR -- `BBMAP_ALIGN` and `FEATURECOUNTS_CDS` both ran with tool defaults, with no way for a user to change ambiguous-read handling.

Why this matters here specifically: the locus-consolidation design from #463 (and any future cross-contig consolidation, #460) assumes each read is exclusively assigned to one ORF at counting time, so summing counts per locus/cluster after the fact never double- or under-counts. That holds for BBMap's default (`best`) and for `random`/`toss` (each produces at most one alignment per read), but breaks under `ambig=all` unless featureCounts is also told to count multi-mapped reads fractionally rather than at full weight per site.

### What changed

- `--bbmap_ambiguous` (`best`/`all`/`random`/`toss`, default `best`) joins the existing `mapping_options` schema group alongside `--bbmap_minid`, and is spliced into `BBMAP_ALIGN`'s `ext.args` as `ambig=${params.bbmap_ambiguous}`.
- `--featurecounts_fraction` (boolean, default `false`) gets its own new `quantification_options` schema group (kept separate from `mapping_options` since it's a featureCounts concern, not a BBMap one -- matches magmap's own precedent here). `FEATURECOUNTS_CDS`'s `ext.args` (previously a plain string) is now a closure appending `-M` unconditionally and `--fraction` only when the flag is set.
- Both params are read directly as `params.x` inside `conf/modules.config`, matching how every other plain scalar param in this pipeline already works (`bbmap_minid`, `min_contig_length`, `prodigal_trainingfile`, etc. -- none of them are threaded through `workflows/metatdenovo.nf`'s `take:` block either).
- Two things intentionally *not* replicated from magmap: its schema omits a `default` for `featurecounts_fraction` even though its `nextflow.config` sets one (schema/config drift, fixed here -- and per `nf-core pipelines schema lint`'s own advice, booleans defaulting to `false` omit the schema `default` key entirely, matching this repo's existing `--save_bam`); and its docs wrongly claim fractional counting is on by default (it isn't, here or there).

### On the `-M` (unconditional) vs `--fraction` (opt-in) split

`-M` only has any effect on reads that already have a secondary alignment in the BAM. Under `ambig=best`/`random`/`toss`, BBMap never produces one, so `-M`'s presence should be a complete no-op by default. This was verified empirically rather than assumed: `tests/default.nf.test` was re-run after this change and still passes clean against its **pre-existing committed snapshot** -- confirming default-params output is genuinely unchanged, not just plausibly so.

### Testing

New `conf/test_bbmap_ambiguous.config` (`bbmap_ambiguous=all`, `featurecounts_fraction=true`, same samplesheet/assembler/orf_caller as the default test profile) + `tests/bbmap_ambiguous.nf.test`, asserting real behaviour rather than just file existence:

- flagstat's secondary-alignment count is `> 0` for every sample (impossible under the default `best`, confirms `all` reached BBMap for real).
- at least one ORF's count is fractional (Subread only emits fractional counts with `--fraction` in effect).
- each sample's summed count never exceeds its total primary mapped reads -- proving fractional weighting keeps totals conserved rather than inflating.

All numbers in the test/snapshot come from a real run (218/69/35417 secondary alignments and 87 fractional-count rows across the three samples), not guessed. Run twice to confirm snapshot stability; `tests/default.nf.test` and `tests/metaeuk.nf.test` (which also exercises `FEATURECOUNTS_CDS` via #463's locus-consolidation codepath) both re-run clean against their existing snapshots.


danilodileo (approved):
LGTM

PR #469, opened 2026-08-25
Add option to keep BBDuk-removed contaminant reads

### LINKED ISSUE #17: Add option to keep removed contaminants
When removing contaminants (e.g. stable RNA) as in #4 we should provide the user with the option of keeping the reads that are filtered out. It is possible to do this with e.g. bbduk by adding `outm` (`outm1` and `outm2` for pair-end reads), where matching reads will be stored, see "[kmer filtering](https://jgi.doe.gov/data-and-tools/bbtools/bb-tools-user-guide/bbduk-guide/)".

danilodileo:
## Summary

Closes #17.

When filtering out contaminants (e.g. stable/rRNA reads) with BBDuk via `--sequence_filter`, matching reads were previously just discarded. This adds `--save_bbduk_removed_fastq`, which uses BBDuk's `outm`/`outm1`/`outm2` options to also keep the reads that were removed (matched the filter reference) instead of only keeping the filtered/clean reads.

- `modules/nf-core/bbmap/bbduk/main.nf`: adds `outm`/`outm1`/`outm2` args (only when a contaminants reference is given), exposed as a new optional `reads_removed` output (`*.matched.fq.gz`) — a distinct extension from the existing `*.fastq.gz` output so the two can't collide under Nextflow's glob matching. Tracked as `modules/nf-core/bbmap/bbduk/bbmap-bbduk.diff` per this repo's vendored-module convention.
- `conf/modules.config`: new `publishDir` rule gated by `--save_bbduk_removed_fastq` (default `false`, matching the existing `--save_bbduk_fastq` pattern).
- `nextflow.config` / `nextflow_schema.json`: new param definition.
- `docs/usage.md`, `docs/output.md`, `CHANGELOG.md`: updated.

## Test plan

- [x] `modules/nf-core/bbmap/bbduk/tests/main.nf.test` extended to snapshot the new `reads_removed` output; full module nf-test suite (6/6) passing with regenerated snapshot.
- [x] `nextflow lint .`, `prek run -a`: clean, no errors/warnings.
- [x] Full end-to-end run (`-profile test,docker --save_bbduk_removed_fastq`): confirmed real `*.matched.fq.gz` files published with expected filtered-read counts.
- [x] Pipeline-level regression check: `tests/default.nf.test` and `tests/single.nf.test` (the only pipeline-level tests exercising BBDuk) pass unchanged against their existing snapshots — confirms the new opt-in output doesn't affect default runs (flag defaults to `false`).


modules/nf-core/bbmap/bbduk/main.nf
```diff
@@ -25,6 +26,7 @@ process BBMAP_BBDUK {
     def raw      = meta.single_end ? "in=${reads[0]}" : "in1=${reads[0]} in2=${reads[1]}"
     def trimmed  = meta.single_end ? "out=${prefix}.fastq.gz" : "out1=${prefix}_1.fastq.gz out2=${prefix}_2.fastq.gz"
     def contaminants_fa = contaminants ? "ref=$contaminants" : ''
+    def matched  = contaminants ? (meta.single_end ? "outm=${prefix}.matched.fq.gz" : "outm1=${prefix}_1.matched.fq.gz outm2=${prefix}_2.matched.fq.gz") : ''
```
erikrikarddaniel: This is the one I'd like your view on before merging.

`outm` is emitted whenever a contaminants reference is given -- it is not gated on `--save_bbduk_removed_fastq`, which only controls `publishDir`. So for anyone running `--sequence_filter` with the flag left at its default `false` (i.e. everyone today), BBDuk now writes a full second copy of the matched reads into `work/` for every sample, and then we throw it away.

On a >90-sample project that is a lot of I/O and scratch space for output nobody asked for.

Since `reads_removed` is already `optional: true`, the output declaration copes fine with the files not existing. So this could move into `ext.args` in `conf/modules.config`, gated on the param, and BBDuk would simply not produce them unless asked:

```groovy
ext.args = { params.save_bbduk_removed_fastq && params.sequence_filter
    ? ( meta.single_end
        ? "outm=${meta.id}.matched.fq.gz"
        : "outm1=${meta.id}_1.matched.fq.gz outm2=${meta.id}_2.matched.fq.gz" )
    : '' }
```

That also keeps the vendored module free of pipeline params, which is the usual nf-core split. Downside is it duplicates the prefix logic in the config and diverges a bit from how you would want it upstream. Happy either way -- but if it stays as is, worth a line in the docs saying the files are always produced and the flag only controls whether they are kept.

modules/nf-core/bbmap/bbduk/main.nf
```diff
@@ -12,7 +12,8 @@ process BBMAP_BBDUK {
     path contaminants
 
     output:
-    tuple val(meta), path('*.fastq.gz'), emit: reads
+    tuple val(meta), path('*.fastq.gz')   , emit: reads
+    tuple val(meta), path('*.matched.fq.gz'), emit: reads_removed, optional: true
     tuple val(meta), path('*.log')     , emit: log
```
erikrikarddaniel: Harshil alignment drifted here -- the new, longer `path('*.matched.fq.gz')` shifted the column, so the three commas no longer line up (3 spaces / 0 / 5). `prek` and prettier will not catch this one.

```suggestion
    tuple val(meta), path('*.fastq.gz')     , emit: reads
    tuple val(meta), path('*.matched.fq.gz'), emit: reads_removed, optional: true
    tuple val(meta), path('*.log')          , emit: log
```

.gitignore
```diff
@@ -16,3 +16,4 @@ null/
 /diamond/
 /gtdb_r220_taxdump/
 /kofamscan/
+.claude/
```
erikrikarddaniel: This looks unrelated to #17 -- it is personal tooling config rather than anything the pipeline produces. Given the nf-core consensus against carrying agent config in pipeline repos, it is probably better in your own global ignore file (`~/.config/git/ignore`, or `git config --global core.excludesfile`), which also means you do not have to re-add it to every repo you work in.

```suggestion
```

modules/nf-core/bbmap/bbduk/tests/main.nf.test
```diff
@@ -82,7 +82,8 @@ nextflow_process {
                 { assert process.success },
                 { assert process.out.reads.get(0).get(1).endsWith("test.trim.fastq.gz") },
                 { assert path(process.out.log.get(0).get(1)).getText().contains("Input is being processed as unpaired")},
-                { assert snapshot(process.out.findAll { key, val -> key.startsWith('versions') }).match() }
+                { assert snapshot(process.out.reads_removed,
```
erikrikarddaniel, line 87: Good -- this is the right test to have extended, and the regenerated snapshot has real md5s for both the single- and paired-end contaminants cases, so the assertion genuinely bites.

Minor gap on the other side of it though: the two *no*-contaminants tests (`sarscov2 - single end fastq - fastq` and its paired-end twin) do not assert anything about `reads_removed`, so the "only produced when a contaminants reference is provided" contract that `meta.yml` now states is not actually pinned. If someone later made `matched` unconditional, those two would still pass.

One line in each would close it:

```groovy
{ assert process.out.reads_removed.size() == 0 },
```

conf/modules.config
```diff
@@ -72,6 +72,12 @@ process {
                 mode: params.publish_dir_mode,
                 enabled: params.save_bbduk_fastq
             ],
+            [
+                path : { "${params.outdir}/bbmap/bbduk/"},
+                pattern: '*.matched.fq.gz',
+                mode: params.publish_dir_mode,
```
erikrikarddaniel, line 79: Minor, and only relevant if a test profile ever sets `--save_bbduk_removed_fastq`: `tests/.nftignore` has `bbmap/bbduk/*.fastq.gz` but not the new `*.matched.fq.gz`, so the matched reads would get md5'd into pipeline snapshots. Harmless today since no profile enables the flag -- just easy to forget later. (I managed to do exactly this to myself on #460 with a different output.)

erikrikarddaniel: The above was Claude: My take, why not publish with the same extension, i.e. `.fastq.gz`? Or is this made difficult by the module?

docs/output.md
```diff
@@ -101,6 +101,7 @@ BBduk is built-in tool from BBmap.
 
 - `bbmap/`
   - `bbduk/*.fastq.gz`: fastq files after removal of sequences specified with `--sequence_filter`. Only saved if you set `--save_bbduk_fastq`.
+  - `bbduk/*.matched.fq.gz`: fastq files with the reads that were removed (matched `--sequence_filter`). Only saved if you set `--save_bbduk_removed_fastq`.
```
erikrikarddaniel, line 104: See my comment above, more consistent to save as `*.matched.fastq.gz`.

modules.json
```diff
@@ -13,7 +13,8 @@
                     "bbmap/bbduk": {
                         "branch": "master",
                         "git_sha": "6d46786420b4d7bc88eba026eb389c0c5535d120",
-                        "installed_by": ["modules"]
+                        "installed_by": ["modules"],
+                        "patch": "modules/nf-core/bbmap/bbduk/bbmap-bbduk.diff"
```
erikrikarddaniel, line 17: As I think Claude will comment elsewhere: It would be good to update the nf-core module. But, let's wait with that and do it with a patch for the time being.

erikrikarddaniel (commented):
Overall good, but Claude found one important thing: We should make sure files are not produced rather than not published. See comment in `main.nf`. Some other comments too.

modules/nf-core/bbmap/bbduk/main.nf
```diff
@@ -33,14 +35,17 @@ process BBMAP_BBDUK {
         threads=$task.cpus \\
         $args \\
         $contaminants_fa \\
+        $matched \\
```
erikrikarddaniel, line 38: Heads-up rather than a request, and there is nothing to change here.

This PR invalidates the `BBMAP_BBDUK` cache for everyone with a run in progress, including runs that never set the flag -- and with BBDuk go the assembly and everything downstream.

I initially thought the cause was this line rendering as a whitespace-only continuation when `matched` is empty, and that folding the newline into the variable would keep the rendered command byte-identical and preserve the cache. That turns out to be wrong, so I am not suggesting it. Nextflow hashes the script *template as written*, not the command it renders to.

Measured, in a minimal pipeline:

- `${a}` changed to `$a`: `.command.sh` byte-identical, task re-executes (`cached=0`).
- a Groovy comment added above the `"""` block: task stays cached (`cached=1`).
- the dev-form line set versus this one, rendering identical `.command.sh`: re-executes.

So any edit to BBDuk's script block invalidates it, and no formulation of this feature avoids that. It only affects *when* this is a good moment to merge -- anyone mid-way through a long run restarts from BBDuk.

modules/nf-core/bbmap/bbduk/tests/nextflow.config
```diff
@@ -3,6 +3,7 @@ process {
     withName: BBMAP_BBDUK {
         ext.args = 'trimq=10 qtrim=r'
         ext.prefix = { "${meta.id}.trim" }
+        ext.save_removed = true
```
erikrikarddaniel, line 6: This makes every test in the file run with the feature on, so the default behaviour -- contaminants given, `save_removed` off, no matched files produced -- is the one case with no test.

It is also the case that motivated the change, and the one every current user gets. One more test with a config that leaves `save_removed` unset, asserting `process.out.reads_removed.size() == 0` with contaminants supplied, would pin it.

erikrikarddaniel (approved):
Small comments remaining, otherwise good!

Issue #472, opened 2026-08-25
Option to annotate only the protein-consolidated representatives


erikrikarddaniel:
## Summary

Add a way to annotate only the protein-consolidated representatives, instead of annotating them
*in addition to* each ORF caller's own ORFs.

Cross-contig protein consolidation (#460) mixes the cluster representatives into `ch_protein` as a
pseudo-caller, so EggNOG, KOfamScan, dbCAN, EUKulele, Diamond and HMMER all run over them alongside
every active caller's own set. With two callers that means annotating **three** protein sets rather
than two, and the third is roughly the size of the union of the other two -- on a full-size
metatranscriptome, 68300 representatives on top of Prokka's and TransDecoder's ORFs. Annotation is
usually the long pole in a real run, so this is close to a doubling of the expensive part.

That was a deliberate choice when #460 was designed: annotating in addition keeps every per-caller
annotation table intact and directly comparable to the consolidated one, which is what you want
when you are still deciding how much to trust consolidation. The cost only becomes the dominant
consideration on large projects, where the consolidated table is the one actually used downstream.

## This is not just a flag

Three decisions have to be made together, which is why it is worth its own issue rather than a
quick addition:

1. **What happens to the per-caller annotation tables.** `<assembly>.<orfcaller>.emapper.tsv.gz`,
   `.kofamscan.tsv.gz`, `.dbcan.tsv.gz`, `.eukulele.taxonomy.tsv.gz` and the HMMER ranks would stop
   being produced for the individual callers. That is the point of the option, but it is a
   user-visible removal, not just a speedup.
2. **How it interacts with `--skip_protein_consolidation`.** The two together are contradictory --
   annotate only a consolidated set that is never produced. It should either error at validation or
   be documented as ignored; silently annotating nothing would be the worst outcome.
3. **What the default should be.** Additive is the safe default and is what #460 ships. Whether it
   stays the default is worth revisiting once there is a real measurement of the annotation cost
   (see below).

There is also a wiring hazard. `MERGE_TABLES` and the `ch_fcs_mergetab_per_caller` join feeding
`COLLECT_STATS` are both keyed on `meta.caller`, and adding a single pseudo-caller to that machinery
in #460 already produced a null-meta crash that had to be fixed with a shared
`ch_counts_per_caller` channel. Removing callers from it needs the same care and its own tests.

## Suggested next step

The full-size test currently running with `--orf_caller prokka,transdecoder` will give a real
[...]

Issue #471, opened 2026-08-25
test_full: writing a downloaded database to an s3:// storeDir fails, aws: command not found


erikrikarddaniel:
## Description of the bug

`-profile test_full` cannot write a downloaded database into an `s3://` `storeDir`, because the
AWS CLI is not available in the environment:

```
.command.run: line 217: aws: command not found
```

The download itself succeeds. In the failing run `CAZy.dmnd`, `dbCAN-sub.hmm` and
`fam-substrate-mapping.tsv` were all fetched, and the failure came afterwards, when Nextflow tried
to copy them into the `storeDir`.

### How the databases are obtained

This is worth stating explicitly, because the parameter names suggest otherwise:

| Tool | Mechanism | In `conf/test_full.config` |
| --- | --- | --- |
| dbCAN | downloads, caches via `storeDir` (`conf/modules.config:409`) | `dbcan_dbpath` -> `s3://` |
| EggNOG | downloads, caches via `storeDir` (`conf/modules.config:353`) | `eggnog_dbpath` -> `s3://` |
| KOfamScan | downloads, caches via `storeDir` (`conf/modules.config:374`) | `kofam_dir` not set -> local `./kofam/` |
| EUKulele | consumes a pre-existing database via `fromPath` | `eukulele_dbpath` -> `s3://` |

So `dbcan_dbpath`/`eggnog_dbpath`/`kofam_dir` are cache **destinations**, not sources. The
databases do not need to be uploaded anywhere -- the pipeline fetches them. Only EUKulele takes a
database as input, and that one is present in the bucket (`test-data/metatdenovo/gtdb_eukulele/`)
and works.

Consistent with this, `s3://ngi-igenomes/test-data/metatdenovo/dbcan/` is empty. That is not a
missing upload: it is a write destination that has never been successfully written to, because the
copy has always failed for the reason above.

### Consequences

- dbCAN and EggNOG both fail this way in the full test. dbCAN is currently skipped in
  `conf/test_full.config` to get the profile running; that is a workaround for the missing CLI, not
  a fix.
- Once the AWS CLI is available, **dbCAN can simply be re-enabled** -- nothing needs uploading.
- **KOfamScan can be enabled at the same time.** It is skipped in the full test today, but it works
  exactly like the other two. Pointing `kofam_dir` at a bucket prefix as well would let it cache
  between runs instead of re-downloading from genome.jp every time.
- Since `awsfulltest` is the workflow that exercises this profile, it is worth checking whether its
[...]

PR #465, opened 2026-08-18
Add MetaEuk as a splice-aware --orf_caller alternative for eukaryotes

erikrikarddaniel:
## PR checklist

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/metatdenovo/tree/master/docs/CONTRIBUTING.md)
- [x] If necessary, also make a PR on the nf-core/metatdenovo _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [x] Make sure your code lints (`nf-core pipelines lint`).
- [x] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [x] Usage Documentation in `docs/usage.md` is updated.
- [x] Output Documentation in `docs/output.md` is updated.
- [x] `CHANGELOG.md` is updated.
- [x] `README.md` is updated (including new tool citations and authors/contributors).

## Description

Addresses #459: adds [MetaEuk](https://github.com/soedinglab/metaeuk) as a new `--orf_caller metaeuk` option, a splice-aware ORF caller for eukaryotes.

Unlike TransDecoder (the existing eukaryote-targeted alternative), MetaEuk aligns contigs against a reference protein database (`--metaeuk_db`, no default -- required whenever `--orf_caller metaeuk` is used) and can call a single gene model spanning an intron, which matters for assemblies mixing intron-containing genomic sequence with spliced transcripts.

Since MetaEuk's native GFF output represents a multi-exon CDS as multiple separate lines with no shared `ID=` (using `Target_ID=`/`TCS_ID=`/`Parent=` instead), a new local module (`FORMAT_METAEUK_GFF`) reformats it into one shared `ID=` per gene, derived from stripping the `_CDS_N` suffix off `TCS_ID`. This lets `featureCounts -g ID -t CDS -F gtf` correctly sum reads across all exons of a multi-exon gene into a single count, instead of splitting them into separate rows. Verified end-to-end against a real run: a 2-exon gene's reads (236) summed correctly into one row spanning both exon coordinates.

Also added:

[...]

PR #479, opened 2026-08-26
Consolidate calls for the same gene across contigs by clustering proteins

### LINKED ISSUE #460: Consolidate ORF/protein calls across multiple gene callers in mixed prokaryote/eukaryote projects

erikrikarddaniel:
## PR checklist

- [ ] Make sure your code lints (`nf-core pipelines lint`).

## Description

Closes #460, the last piece of the arc after #465, #466, #467, #468.

Coordinates cannot recognise the same gene called on two *different* contigs. This clusters the loci' proteins with MMseqs2 (official `mmseqs_fasta_cluster` subworkflow) and sums level-2 counts per cluster, adding `<assembly>.protein_consolidate_<pct>.counts.tsv.gz` alongside the per-caller and `locus_consolidate` tables. Reads are not remapped. `--cluster_min_seq_id` (0.99), `--cluster_coverage` (0.8), `--skip_protein_consolidation`.

Annotation is **additive**: representatives join `ch_protein` as a pseudo-caller, so every per-caller table is unchanged. Annotating *only* representatives is #472.

### For a reviewer

- **Changes an already-merged PR's output.** #467 used `bedtools merge -s`, which cannot tell two callers agreeing on one gene from one caller calling two adjacent genes. On the default test data it fused **515 of 3470 Prodigal ORFs (14.8%)** into 238 loci, summing counts across distinct genes. Replaced by a caller-aware sweep; now 3470 -> 3470, zero fusions.
- **Levels 2 and 3 were not reproducible.** Equal-start order came from `collectFile(sort: true)` ordering BEDs by work-dir path, so a re-run flipped it: 68829 / 68826 / 68828 loci on full-size data. The module now sorts its own input on a total order (`LC_ALL=C`).
- **Two unreleased bugs ride along.** MetaEuk's FASTA and GFF disagreed on the ID (fields 1,2,3,**7**), so every annotation `inner_join(by = 'orf')` returned zero rows for MetaEuk. And with several callers, eggnog-mapper annotated only the first: `.combine()` demoted the database value channel to a one-item queue channel. Neither was reachable by any test profile.
- **Two local modules renamed** to nf-core convention (`format/locusconsolidate`, `collect/locusconsolidate`); the published table name is unchanged.
- **`conf/test.config` keeps consolidation on** rather than adding a profile -- it is on by default, and single-caller is exactly where degeneracy should hold.
- **dbCAN skipped in `test_full`**, see #471 (`storeDir` on `s3://` needs `aws`), unrelated to this PR.

### Validation

Full-size: 68829 loci -> 68300 clusters; 476 clusters merged >1 locus, covering 1005 loci (1.5%), 315 cross-caller. Level-3 counts equal the sum of member level-2 counts for every cluster and sample; no locus in two clusters; TPM recomputed, sums to 1e6.

10 of 12 profiles regenerated locally. `test_diamond`/`test_kofamscan` were validated on a machine with those databases; nothing since can affect them (single-caller, consolidation skipped, no eggnog). New module tests for all five new/changed local modules, inputs built inline.

`CHANGELOG.md` `#NN` placeholders need this PR's number.

Written with help from Claude Code.


danilodileo (approved):
As in the best of my knowledge, this looks ok. We already discussed that it would be probably good to make shorter PRs. Even though it's tempting to fix and make changes all in one goal, sometimes it is hard then to review. I see that many tests are performed and everything looks clean to my eyes. I can approve it and we will see how it performs on real data. good work!

erikrikarddaniel:
> As in the best of my knowledge, this looks ok. We already discussed that it would be probably good to make shorter PRs. Even though it's tempting to fix and make changes all in one goal, sometimes it is hard then to review. I see that many tests are performed and everything looks clean to my eyes. I can approve it and we will see how it performs on real data. good work!

I completely agree. Parts of this PR belong together in a way that makes it difficult to split into different PRs, but there was certainly room for more PRs. And Claude has reminded me about this rule a couple of time, but I pushed on, so, mea culpa.

I will start a run with this code soon, probably tomorrow, and then do a full "sanity check". Me and Claude also went quite deep into the output from `test_full` to check that the whole sequence of PRs make sense.

Issue #477, opened 2026-08-26
TRANSDECODER_PREDICT breaks -resume and fails on every re-execution


erikrikarddaniel:
## What

`TRANSDECODER_PREDICT` can never be resumed, and its re-execution always fails.
Reported upstream as [nf-core/modules#12799](https://github.com/nf-core/modules/issues/12799), which has the full analysis and a reproduction.

Short version: `TransDecoder.Predict` writes its `__checkpoints_TDpredict` directory into the LongOrfs directory it receives as a staged input, which belongs to the `TRANSDECODER_LONGORF` task.
So the task modifies its own input and stops being cacheable, and when Nextflow re-executes it the task directory is wiped while the checkpoints -- living in the LongOrfs task's directory -- are not.
`TransDecoder.Predict` then skips every step, including the ones that write the outputs, and exits 0 with nothing produced.

Reproduced on this pipeline's own smallest profile, roughly two minutes per run:

```bash
nextflow run . -profile test,docker --outdir out --orf_caller transdecoder
# [SUCCESS] completed=63 failed=0 cached=0

nextflow run . -profile test,docker --outdir out --orf_caller transdecoder -resume
# TRANSDECODER_LONGORF   CACHED
# TRANSDECODER_PREDICT   FAILED   (Missing output file(s) '*.transdecoder.pep', exit status 0)
```

## Why this needs fixing before 2.0

TransDecoder is one of the two ORF callers the full-size test uses, and the one recommended for eukaryotic data.
Every long run on real data eventually gets interrupted, and every resume that reaches this task fails -- so in practice a large TransDecoder run cannot be restarted, only redone.
On a multi-sample metagenome assembly that is hours of recomputation, and with `errorStrategy 'retry'` it turns a transient failure into a permanent one.

## Plan

Give the upstream maintainers a few days to comment or fix it.
If nothing has moved by then, do it ourselves: open the PR against nf-core/modules and, if it has not been merged when 2.0 approaches, carry it locally with `nf-core modules patch transdecoder/predict` so the release is not blocked on upstream timing.

Fix options, from the upstream issue, in order of preference:

1. Give Predict a private writable view of the LongOrfs directory -- a directory of symlinks to each file, so reads follow the links and everything Predict creates is local. No extra disk.
2. `cp -rL` the whole directory. Certainly correct, but duplicates the LongOrfs output, which is large on metagenome assemblies.
[...]

PR #480, opened 2026-08-27
Update transdecoder/predict to fix -resume failure

erikrikarddaniel:
## PR checklist

## Description

Closes #477.

`TRANSDECODER_PREDICT` could never resume, and its re-execution always failed. Full analysis and reproduction in #477 and the upstream report, [nf-core/modules#12799](https://github.com/nf-core/modules/issues/12799).

Short version: `TransDecoder.Predict` writes its checkpoint directory into the LongOrfs directory it receives as a staged input, which belongs to a different task's (`TRANSDECODER_LONGORF`'s) work directory. So the task mutates its own input and stops being cacheable, and when Nextflow re-executes it the task directory is wiped while the checkpoints -- living in the LongOrfs task's directory -- are not. `TransDecoder.Predict` then skips every step, including the ones that write the outputs, and exits 0 with nothing produced.

This PR pulls in the fix, merged upstream in [nf-core/modules#12829](https://github.com/nf-core/modules/pull/12829): `nf-core modules update transdecoder/predict --no-preview`. The module now gives Predict a private, writable directory of symlinks to the same files, instead of writing through the staged symlink into another task's directory.

### Testing

- `tests/metaeuk_transdecoder.nf.test` passes unmodified against the existing committed snapshot -- output content is unchanged.
- Reproduced the bug and verified the fix directly against this pipeline, not just the module in isolation: ran `-profile test,docker --outdir <dir> --orf_caller transdecoder`, then the same command with `-resume`.
  - **Before this fix:** `TRANSDECODER_PREDICT` fails on resume with `Missing output file(s) '*.transdecoder.pep'`, exit status 0 -- exactly as described in #477.
  - **After this fix:** the trace report shows `TRANSDECODER:TRANSDECODER_PREDICT` as `CACHED`, and the whole pipeline resumes cleanly.
- `nf-core pipelines lint` and `prek run -a`: clean (one pre-existing, unrelated bioconda-version warning on the module itself, 5.7.1 -> 6.0.0 available, left untouched -- out of scope for this fix).

Isolated from #479 (the in-review protein-consolidation PR) -- branched off `dev` rather than `issue-460-protein-consolidation`, and touches only the vendored `transdecoder/predict` module, `modules.json`, and the changelog, so it should merge independently without conflict.

Generated by Claude


piplus2 (changes requested):
Thanks for tracking this down — the diagnosis is right and I verified it against the TransDecoder 5.7.1 sources: `TransDecoder.Predict` sets `$checkpoints_dir = "$workdir/__checkpoints_TDpredict"` with `$workdir = "$output_dir/<fasta>.transdecoder_dir"`, so it really does write into the staged LongOrfs directory. I also checked that `Predict` never writes *through* the new symlinks — it only reads `longest_orfs.{cds,gff3,pep}` and `base_freqs.dat`, and every path it writes (`hexamer.scores`, `longest_orfs.cds.*`, `start_refinement.*`, `__checkpoints_TDpredict/`) is a new name. So the symlink-farm approach does fully contain the mutation, and nothing leaks back into `TRANSDECODER_LONGORF`'s work directory. 

The sync itself is clean: all five vendored files byte-match upstream at the new sha `a990e410159e3bc9cebfd0db69982a77c9954c63`, and the `modules.json` bump and CHANGELOG entry are both accurate.

Two things on the shell block before this merges:

### `rm $fold` assumes the input is staged as a symlink (blocking)

```
real_fold=$(readlink -f $fold)
rm $fold
mkdir $fold
ln -s "$real_fold"/* $fold/
```

`$fold` is only a symlink under local/shared-filesystem staging. On AWS/Google/Azure Batch without Fusion, Nextflow's `nxf_stage()` downloads directory inputs (`aws s3 cp --recursive`), so `$fold` is a real directory — and the same happens whenever a user or institutional config sets `process.stageInMode = 'copy'`. In that case `rm` fails with `rm: cannot remove '...': Is a directory`, and since we set

```groovy
process.shell = ["bash", "-C", "-e", "-u", "-o", "pipefail"]
```

in `nextflow.config`, the task aborts right there. That turns a `-resume` annoyance into a hard failure on every cloud run using `--orf_caller transdecoder`.

Worth noting the obvious repair is also wrong: `rm -r $fold` would delete the very copy `real_fold` points at, leaving the `ln -s` with nothing to link. A staging-mode-agnostic version handles both cases, since `mv` moves the link itself for a symlink and renames for a real directory:

```bash
mv $fold ${fold}.staged
real_fold=$(readlink -f ${fold}.staged)
mkdir $fold
ln -s "$real_fold"/* $fold/
```

We won't catch this in CI as things stand — `conf/test.config` pins `prodigal` and `conf/test_full.config` pins `prokka`, so no AWS-executed profile exercises this module.

### Glob without `dotglob` / `nullglob` (minor)

`ln -s "$real_fold"/* $fold/` silently skips any dot-prefixed entry, and if the glob matches nothing it still succeeds and creates a dangling link literally named `*` — so the failure surfaces later as an opaque TransDecoder "cannot find longest_orfs.cds" instead of here. It's latent today, since 5.7.1 writes no dotfiles into `<fasta>.transdecoder_dir`, but you mention the pending 5.7.1 → 6.0.0 bioconda bump in the description, and a hidden checkpoint marker in a future release would break this quietly rather than loudly. A `shopt -s dotglob nullglob` before the `ln` covers it.

Since this is a clean `nf-core modules update` sync (all five vendored files byte-match upstream at the new sha, no `.diff`), both of these are really upstream bugs in `nf-core/modules`. Probably worth raising them there so the fix lands in the module itself — otherwise we either carry a local patch that a future `nf-core modules update` will flag, or we ship the cloud-staging failure. I'd lean towards fixing upstream and re-syncing here, but if that's slow, a local patch for the `rm` is worth it given it's a hard failure rather than a resume annoyance.

piplus2 (approved):
LGTM

erikrikarddaniel:
Thanks @piplus2 -- hard work again! :-)

PR #483, opened 2026-08-28
Keep COLLECT_FEATURECOUNTS generic by extracting Transdecoder cds. prefix stripping

erikrikarddaniel:
## PR checklist

## Description

Part of the cross-repo summary-table-generation consolidation tracked in nf-core/magmap#237: the goal is to converge magmap's and metatdenovo's `COLLECT_FEATURECOUNTS` modules on identical, generic logic so the two can eventually share a real `nf-core/modules` component. magmap's side is already done (magmap#238): it moved its genome-accession-join special-case out of the generic aggregator into a small local post-processing module, `TIDYVERSE_JOINFEATURECOUNTSACCNO`.

This PR does the equivalent for metatdenovo's own special-case: `COLLECT_FEATURECOUNTS` stripped Transdecoder's `cds.` ORF-ID prefix inline (Transdecoder adds it in its GFF but not its FASTA, so counts tables need it removed to join against FASTA-derived IDs used by the annotation subworkflows). That's now a separate local module, `TIDYVERSE_STRIPCDSPREFIX`, run as a post-processing step right after `COLLECT_FEATURECOUNTS`. `COLLECT_FEATURECOUNTS` itself no longer has any Transdecoder-specific knowledge.

No change to the content of any output table — verified with real Docker runs of the three pipeline tests that exercise the Transdecoder caller (`megahit_transdecoder`, `megahit_prokka_transdecoder`, `metaeuk_transdecoder`); the only snapshot diff in all three is the new module's entry in the collated software-versions YAML.

Note `COLLECT_LOCUSCONSOLIDATE` and `COLLECT_PROTEINCONSOLIDATE` (added since the original #237 plan was written) carry their own independent copies of the same `cds.` stripping logic, operating on the raw `*.featureCounts.tsv` files directly rather than on `COLLECT_FEATURECOUNTS`'s output. Those are metatdenovo-specific consolidation modules with no magmap equivalent, so they're out of scope here — left untouched.

Generated by Claude

### DIFF
--- a/modules/local/collect/featurecounts/main.nf
+++ b/modules/local/collect/featurecounts/main.nf
@@ -52,8 +52,6 @@ process COLLECT_FEATURECOUNTS {
             )
         ) %>%
         tidyr::unnest(d) %>%
-        # Transdecoder appends "cds." to ORF IDs in the gff file, but does not in the fasta file. Remove to make compatible between tables.
-        mutate(orf = str_remove(orf, '^cds\\\\.')) %>%
         select(-f) %>%
         write_tsv("${prefix}.counts.tsv.gz")
 
--- a/modules/local/collect/featurecounts/main.nf
+++ b/modules/local/collect/featurecounts/main.nf
@@ -45,11 +45,7 @@ process COLLECT_FEATURECOUNTS {
                         ) %>%
                         rename(orf = Geneid, chr = Chr, start = Start, end = End, strand = Strand, length = Length) %>%
                         group_by(sample) %>%
-                        # Rounded so a single caller's table is byte-comparable to COLLECT_LOCUSCONSOLIDATE's/
-                        # COLLECT_PROTEINCONSOLIDATE's independently-computed tpm for the same counts --
-                        # data.table/dtplyr here vs plain dplyr there can otherwise round the last
-                        # significant digit differently for the same mathematical value.
-                        mutate(tpm = round(r/sum(r) * 1e6, 6)) %>% ungroup() %>%
+                        mutate(tpm = round(r/sum(r) * 1e6, 6)) %>% ungroup() %>%  # round so consistent
                         select(-r) %>%
                         as_tibble()
                 }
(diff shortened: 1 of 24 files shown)


modules/local/collect/featurecounts/main.nf
Joon-Klaps, line 1: would remove these comments

modules/local/collect/locusconsolidate/main.nf
```diff
@@ -51,7 +51,10 @@ process COLLECT_LOCUSCONSOLIDATE {
                         ) %>%
                         rename(orf = Geneid, chr = Chr, start = Start, end = End, strand = Strand, length = Length) %>%
                         group_by(sample) %>%
-                        mutate(tpm = r/sum(r) * 1e6) %>% ungroup() %>%
+                        # Rounded so a single caller's own table (from COLLECT_FEATURECOUNTS's data.table/
+                        # dtplyr computation) is byte-comparable to this one -- the unrounded value can
+                        # otherwise round the last significant digit differently between the two backends.
+                        mutate(tpm = round(r/sum(r) * 1e6, 6)) %>% ungroup() %>%
```
Joon-Klaps: Remove or simplify no?
```suggestion
                        mutate(tpm = round(r/sum(r) * 1e6, 6)) %>% ungroup() %>%  # round so consistent
```

modules/local/collect/proteinconsolidate/main.nf
```diff
@@ -102,7 +102,9 @@ process COLLECT_PROTEINCONSOLIDATE {
         summarise(count = sum(count), .groups = 'drop') %>%
         left_join(cluster_attrs, by = 'cluster') %>%
         group_by(sample) %>%
-        mutate(tpm = (count/length)/sum(count/length) * 1e6) %>%
+        # Rounded for consistency with COLLECT_FEATURECOUNTS/COLLECT_LOCUSCONSOLIDATE, which round for
+        # the same reason: keeps output stable regardless of which R backend computed the division.
```
Joon-Klaps: same

Joon-Klaps (approved):

PR #492, opened 2026-09-02
Batch MetaEuk and TransDecoder like Prokka already is

erikrikarddaniel:
## PR checklist

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/metatdenovo/tree/master/docs/CONTRIBUTING.md)
- [ ] If necessary, also make a PR on the nf-core/metatdenovo _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [ ] Make sure your code lints (`nf-core pipelines lint`).
- [ ] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [x] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [x] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).

## Description

Batches MetaEuk's ORF calling and TransDecoder's LongOrfs/Predict on split input (assembly contigs for MetaEuk, transcripts for TransDecoder), the same `splitFasta`-based mechanism `PROKKA_SUBSETS` already uses. Addresses part of #486.

Motivated by a live large-scale run (~82M contigs) losing a multi-hour, whole-assembly `METAEUK_EASYPREDICT`/`TRANSDECODER_LONGORF`/`TRANSDECODER_PREDICT` task to a single failure (including hitting a genuine Nextflow engine bug, `IllegalThreadStateException` aborting the whole session). Batching bounds how much work a killed/failed batch costs, and for MetaEuk also bounds `extractorfs`'s peak memory, which scales with how many contigs it's handed at once.

Note: `TransDecoder.Predict` self-trains its coding/noncoding model per batch (confirmed via `--help`, no option to reuse a model trained elsewhere), so batching changes its statistical behaviour slightly versus one model over the whole assembly. A very large `--transdecoder_batchsize` recovers the unbatched behaviour exactly for anyone who needs that.

The other half of #486 (batching TransRate) is superseded by #491, which replaces TransRate with QUAST instead of parallelizing it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #486, opened 2026-08-30
Batch METAEUK and TRANSRATE like PROKKA_SUBSETS, for resumability and MetaEuk's OOM risk


erikrikarddaniel:
## Background

Hit this concretely on a live large-dataset run (2026-08-30): after ~30 hours and 1852 successful tasks, a Nextflow engine-internal race (`IllegalThreadStateException: process hasn't exited` in `LocalTaskHandler.checkIfCompleted`, unrelated to any of our code) hard-aborted the whole session while `METAEUK_EASYPREDICT` and `TRANSRATE` were both still running.
Because neither task had ever recorded an exit code, `-resume` cannot reuse any of that in-flight work — both restart from scratch.
For `METAEUK_EASYPREDICT` on an 82-million-contig assembly, that single task alone can run for many hours (and separately hit a real OOM earlier in the same investigation, see #485's discussion history in Slack/PR context).

We already solve exactly this problem for Prokka: `subworkflows/local/prokka/subsets/main.nf` (`PROKKA_SUBSETS`) splits the assembly FASTA into `params.prokka_batchsize`-sized chunks with Nextflow's own `.splitFasta(size: batchsize, file: true)` operator, runs `PROKKA` once per chunk, then concatenates the per-chunk GFF/FAA/FFN with `FIND_CONCATENATE`. Originally added because Prokka itself failed outright on large contigs; it turns out to be equally valuable as a resumability boundary — losing one chunk only costs re-running that chunk, not the whole assembly.

Proposal: apply the same pattern to `METAEUK_EASYPREDICT` and to `TRANSRATE`.

## `METAEUK_SUBSETS`

A close-to-direct port of `PROKKA_SUBSETS`. MetaEuk's splice-aware ORF/gene calling is inherently per-contig — it never needs cross-contig context — so splitting the assembly the same way and concatenating the per-batch GFF and protein FASTA outputs should be safe with no special-casing.

This is a double win, not just resumability: `extractorfs`'s peak memory scales with however many contigs it's handed at once (this is what OOM'd on the full 82M-contig assembly). Batching directly shrinks that peak per task, likely removing the need for the large explicit memory ceiling we ended up requesting on the unbatched run.

## `TRANSRATE_SUBSETS`

Less of a direct port — `TRANSRATE` (`modules/nf-core/transrate/main.nf`) emits two outputs with different aggregation properties:

[...]

PR #495, opened 2026-09-06
Fix TRANSDECODER_PREDICT pairing the wrong batch's fasta with another batch's LongOrfs output

erikrikarddaniel:
## PR checklist

- [ ] `CHANGELOG.md` is updated. — Intentionally skipped: the batched TransDecoder feature this bug lives in (#492) hasn't shipped in a release yet, so this fix folds into that same not-yet-released feature rather than getting its own user-facing entry.

## Description

`TRANSDECODER_LONGORF.out.folder` carries no `meta`/id, so `TRANSDECODER_PREDICT` paired it with `ch_batches` purely by channel consumption order. That order isn't guaranteed to match submission order once batches run concurrently: a smaller/faster batch's LongOrfs can finish before a bigger, earlier-submitted batch's, silently swapping which fasta gets predicted against which batch's LongOrfs directory.

Hit on a real 82M-contig production run, where `contigs.1`'s fasta was paired with `contigs.78`'s LongOrfs directory and crashed with `cannot find directory ... contigs.1.fa.transdecoder_dir`. Every actual mismatch fails loudly this way (TransDecoder.Predict's expected directory name is derived from its own fasta argument, not from whatever directory it's actually handed), so this manifests as crashes rather than silently wrong biology — but it does mean any batch that happens to race can fail for a batch size where it previously worked.

Added a failing-first `nf-test` for the `TRANSDECODER` subworkflow that forces the same race deterministically with two batches of very different size (so the small one reliably finishes LongOrfs first, regardless of the runner's core count), confirmed it reproduces the exact same "cannot find directory" crash, then fixed it by joining `ch_batches` and `TRANSDECODER_LONGORF.out.folder` explicitly on id — recovered from the folder's own parent directory name — instead of relying on positional pairing.

Addresses part of #486, in the same batching work introduced by #492 (not yet released).

---

### DIFF
--- a/subworkflows/local/transdecoder/main.nf
+++ b/subworkflows/local/transdecoder/main.nf
@@ -37,7 +37,22 @@ workflow TRANSDECODER {
 
     TRANSDECODER_LONGORF ( ch_batches )
 
-    TRANSDECODER_PREDICT ( ch_batches, TRANSDECODER_LONGORF.out.folder )
+    // TRANSDECODER_LONGORF.out.folder carries no meta, so pairing it with ch_batches
+    // positionally (as two separate process inputs) relies on both channels emitting in the
+    // same order -- which isn't guaranteed once batches run concurrently: a smaller/faster
+    // batch's LongOrfs can finish before an earlier, bigger batch's, silently swapping which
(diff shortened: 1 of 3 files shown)


danilodileo (approved):
👍

Issue #499, opened 2026-09-07
No test coverage of annotation tools run against protein-consolidated cluster representatives


erikrikarddaniel:
## Background

While implementing `--annotate_only_consolidated` it turned out **no existing test feeds any annotation tool the protein-consolidated cluster representatives** (`protein_consolidate_<identity>` caller, produced by `#460`/`#479`).

Checked across `dev`, `#494` (`user-orfs-additive`), and `#495` (`fix-transdecoder-predict-order`) -- all three have the identical set of 12 `tests/*.nf.test` files, so this isn't a branch-fragmentation artifact, it's a real gap. The only test config that combines (>1 ORF caller) + (protein consolidation on) + (an annotation tool enabled) is `conf/test_full.config` (`prokka,transdecoder`, `skip_eggnog=false`, `skip_eukulele=false`, plus `hmmfiles` for HMMCLASSIFY) -- and it has no corresponding `.nf.test` file, so it has never actually run in CI.

Every other multi-caller test either explicitly sets `--skip_protein_consolidation` (`test_megahit_prokka_prodigal`, `test_megahit_prokka_transdecoder`) or has no annotation tool enabled at all (`test_metaeuk_transdecoder`).

## Why it matters

The annotation subworkflows each join their own output back onto per-caller feature counts by `meta.caller` (e.g. `KOFAMSCAN_SUM`'s `.join()` on `meta.caller` in `subworkflows/local/kofamscan/main.nf`, similarly in EGGNOG/DBCAN/EUKULELE). None of these joins have ever actually seen a `protein_consolidate_<identity>`-shaped caller id in a test. A bug specific to that caller name/shape (e.g. an id-format assumption, or a join failing to find the corresponding `COLLECT_PROTEINCONSOLIDATE` counts) could ship silently.

## Partially addressed

The new `tests/annotate_only_consolidated.nf.test` (part of the `--annotate_only_consolidated` PR) closes this gap for **HMMCLASSIFY** specifically: both its default (`true`) and `--annotate_only_consolidated false` variants assert `summary_tables/megahit.protein_consolidate_99.hmmrank.tsv.gz` is produced against 2 active ORF callers (`prokka,prodigal`) with consolidation on.

## Still open

[...]

PR #501, opened 2026-09-08
Add --annotate_only_consolidated to skip redundant per-caller annotation

erikrikarddaniel:
## PR checklist

## Description

With N active `--orf_caller`/`--user_orfs` sources plus protein consolidation, every annotation subworkflow (EGGNOG/KOFAMSCAN/DBCAN/EUKULELE/DIAMOND_TAXONOMY/HMMCLASSIFY) used to run on each individual source's full protein set *as well as* on the cluster representatives -- up to N-fold repeated annotation of genes multiple sources agree on. This was making annotation take a very long time on a real 82M-contig, multi-source production run.

`--annotate_only_consolidated` (new param, default `true`) restricts those subworkflows to just the cluster representatives once consolidation has run. It's a channel `.filter()` right after the existing protein-consolidation block, so it needs no changes to any downstream consumer.

It's a silent no-op (no warning) in two cases, both by explicit design:
- consolidation didn't run (`--skip_protein_consolidation`, or no ORF source active at all)
- only a single ORF source is active -- protein consolidation can still collapse same-caller cross-contig duplicates in that case, but forcing a single-caller run's annotation output to be renamed/reshaped under `protein_consolidate_*` for that marginal dedup isn't worth the default-behavior change

Verified live on a real 82M-contig, 3-ORF-source production run: caching held across the resume, and annotation correctly waited for clustering before running -- no annotation subworkflow saw any input until `protein_consolidate_*` representatives existed.

**Test coverage gap found while building this:** no existing test (checked across `dev` and every in-flight branch at the time) ever fed a protein-consolidated cluster representative to *any* annotation tool -- `conf/test_full.config` is the only config that would, and it has no `.nf-test` file. Filed as [#499](https://github.com/nf-core/metatdenovo/issues/499), narrowed in a follow-up comment to KOFAMSCAN/DBCAN as the realistic near-term additions (both already have small, CI-cheap test databases; EGGNOG/EUKULELE/DIAMOND_TAXONOMY's test setups are all too heavy to combine with 2 callers + consolidation in CI). This PR's own new test (`tests/annotate_only_consolidated.nf.test`) closes the gap for HMMCLASSIFY, with two `test()` blocks -- default `true` and explicit `--annotate_only_consolidated false` -- asserting presence/absence of the per-source vs. consolidated `hmmrank.tsv.gz` outputs in each direction.


workflows/metatdenovo.nf
```diff
@@ -720,6 +720,16 @@ workflow METATDENOVO {
         ch_protein = ch_protein.mix(SEQKIT_GREP.out.filter)
     }
 
+    // Restrict annotation (everything below still reading plain ch_protein) to the cluster
```
jonasscheid, line 723: Avoid chat context in your comments in codebase. I know Opus 5 tries hard not to :D (non-blocking)

jonasscheid (approved):

PR #494, opened 2026-09-02
Make user-supplied ORFs additive with --orf_caller, support multiple named sets

erikrikarddaniel:
## PR checklist

- [x] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests! — verified manually end-to-end on real hardware (see below), no automated `nf-test` coverage added yet.
- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/metatdenovo/tree/master/docs/CONTRIBUTING.md) — n/a, no new tool.
- [ ] If necessary, also make a PR on the nf-core/metatdenovo _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository. — n/a, no new test data.
- [x] Make sure your code lints (`nf-core pipelines lint`).
- [ ] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [x] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated. — n/a, no new output files; existing output docs already describe consolidation outputs that user-supplied ORFs now participate in.
- [x] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors). — n/a, no new tool.

## Description

`--user_orfs_gff`/`--user_orfs_faa` used to be mutually exclusive with `--orf_caller`, and only
one user-supplied ORF set could be provided per run. This PR removes both restrictions:

- `--orf_caller`, `--user_orfs_gff`+`--user_orfs_faa`, and a new `--user_orfs` CSV are now all
  additive — any combination can be used together, and at least one is required.
- The new `--user_orfs` CSV (`name,gff,faa` columns, schema `assets/schema_user_orfs.json`,
  following the existing `--diamond_dbs` precedent) lets a user supply any number of named
  ORF sets in one run, not just one.
- Every active name (built-in callers, CSV row names, `--user_orfs_name`) must be unique and
  can't collide with names the pipeline reserves for itself (`locus_consolidate`,
  `protein_consolidate_<pct>`) — checked up front with a clear `error()` rather than failing
  downstream.
- User-supplied ORF sets are now mixed into the pipeline exactly like another built-in caller,
  so they participate in locus consolidation, protein consolidation, and feature counting.
[...]

PR #502, opened 2026-09-08
Auto-download/build the MetaEuk reference database, matching KofamScan

erikrikarddaniel:
## PR checklist

## Description

`--metaeuk_db` used to require a pre-built database path with no default and no download support -- the user had to run MetaEuk's own `metaeuk databases <name> <outdir> tmp` outside the pipeline first. This adds a new `METAEUK_DOWNLOAD` module, wired into `subworkflows/local/metaeuk/`, that auto-downloads and builds a database (`--metaeuk_db_name`, default `UniRef50`, cached under `--metaeuk_db_dir`) when `--metaeuk_db` isn't set -- matching KofamScan's existing `KOFAMSCAN_DOWNLOAD` pattern. `--metaeuk_db` remains as an escape hatch for a database the user already has.

Both the default (`UniRef50`) and an alternate (`UniProtKB/Swiss-Prot`) database were confirmed to download and build correctly on real hardware. `PDB` is documented as known-bad (its hardcoded upstream host, `ftp.wwpdb.org`, doesn't resolve -- confirmed via docker, singularity, and a bare `wget`, not a pipeline bug).

Also folded in, found while testing this on a real run reaching `COLLECT_PROTEINCONSOLIDATE` for the first time (unrelated to this branch, but tiny and not worth a separate PR through the review queue): `COLLECT_LOCUSCONSOLIDATE`/`COLLECT_PROTEINCONSOLIDATE` read their gzipped provenance file via `fread(cmd = "zcat ...")` instead of a bare `fread('file.gz', ...)`, which needs the `R.utils` package -- one of the shared Wave-built R containers had it in its Docker tag but not its pinned Singularity blob. Matches the `prokka/gff2tsv` module's existing `zgrep`-based pattern.

**Rebased onto current `dev`** (this work predates the whole multi-ORF-caller/consolidation rework -- #494/#495/#500/#501 and everything before them). The real conflict was in `subworkflows/local/metaeuk/main.nf`: this branch's original `METAEUK` subworkflow called `METAEUK_EASYPREDICT` directly on the whole assembly, while `dev`'s now-merged MetaEuk batching (#492) splits it into per-contig-size batches first. Merged both: batches are built first (unchanged from `dev`), then the batch fasta is handed to whichever database resolution path (`db_path` vs. `db_name`/auto-download) this PR adds.

## Testing

- `modules/local/metaeuk/download/tests/main.nf.test` (2 stub tests: default db name, and a name containing `/` which needs special handling for `db_dir`)
- `tests/metaeuk.nf.test`/`tests/metaeuk_transdecoder.nf.test` (both exercise the rebased subworkflow's `db_path`-set branch end to end)
- `nf-core pipelines lint` clean


docs/usage.md
```diff
@@ -140,22 +140,38 @@ For eukaryotic genes, we recommend users to use Transdecoder (`--orf_caller tran
...
+`--metaeuk_db_name` (default `UniRef50`) picks which database to build.
 
-- `UniRef50`/`UniRef90`/`UniRef100`
-- `UniProtKB/Swiss-Prot`
-- `GTDB`
-- `NR`
+The database itself, and the URL it's fetched from, are entirely MetaEuk's own -- the pipeline doesn't maintain a separate mapping of names to URLs, only the list of names it allows `--metaeuk_db_name` to be.
+That list is restricted to the amino-acid databases `metaeuk databases -h` offers (its nucleotide and profile entries aren't valid MetaEuk homology references), but since the actual download depends on MetaEuk's own logic and a third-party host we don't control, we can only vouch for the ones we've actually confirmed downloadable ourselves:
+
+| Name                   | Confirmed downloadable | Notes                                                                                                                                                         |
```
danilodileo, line 151: Isn't better to test all of these databases before merging this PR?

danilodileo (approved):
LGTM, I think we should give more info regarding which db can be downloaded and for which db the download is broken.

docs/usage.md
erikrikarddaniel: They take *forever*. Even the Uniref50 took several hours.

erikrikarddaniel:
> LGTM, I think we should give more info regarding which db can be downloaded and for which db the download is broken.

In principle I agree. In practice it's difficult to test because of the sizes and since it's not our responsibility I think we can keep it as is. We're at least clear with what we *know* worked at a specific time.

Issue #478, opened 2026-08-26
Adopt typed parameters so booleans can be set from the command line


erikrikarddaniel:
## The problem

A boolean parameter cannot be turned **off** from the command line.
`--skip_dbcan false` and `--skip_dbcan=false` both arrive as the *string* `'false'`, which is truthy in Groovy, so `if ( ! params.skip_dbcan )` stays false and the step remains skipped.
There is no warning, and `"type": "boolean"` in `nextflow_schema.json` does not coerce it.

The bare form `--skip_dbcan` correctly sets `true`, so the asymmetry is easy to miss: turning a step *on* from the command line simply does not work.

This is not hypothetical.
The full-size run of 2026-08-26 was launched with `--skip_dbcan false --skip_kofamscan false`, and `pipeline_info/params_2026-08-26_08-34-24.json` recorded:

```json
"skip_dbcan": "false",
"skip_kofamscan": "false",
"skip_eggnog": false,
"skip_eukulele": false
```

The quoting is the whole story: the two set on the command line are strings, the two coming from the config are booleans.
No dbCAN or KOfamScan task appears anywhere in that run's execution trace.
The run took most of a day and silently produced neither annotation.

A related failure mode is that a mistyped parameter name is accepted in silence -- `--skip_dbcna false` sets a new parameter nobody reads.

## Typed parameters fix both

Nextflow's typed `params` block coerces the value against the declared type:

```groovy
params {
    skip_dbcan: Boolean = false
}
```

Measured on Nextflow 26.04.6:

| invocation | resulting value | effect |
| --- | --- | --- |
| `--skip_dbcan false` | `false` (Boolean) | step **runs** |
| `--skip_dbcan` | `true` (Boolean) | step skipped |
| `--skip_dbcan true` | `true` (Boolean) | step skipped |
| `--skip_dbcan maybe` | -- | hard error: ``Parameter `skip_dbcan` with type Boolean cannot be assigned to maybe [String]`` |

The last row matters as much as the first: wrong values become errors instead of surprises.

## Cost: it requires Nextflow >= 26.04

Under `NXF_VER=25.10.4` the typed `params` block does not compile at all (`ERROR ~ Script compilation error`), and `nextflow.config` currently declares `nextflowVersion = '!>=25.10.4'`.

Adopting this therefore means raising the minimum Nextflow version, which is a major-release decision -- hence 2.0 rather than a patch.

## Scope

[...]

PR #503, opened 2026-09-09
Coerce CLI-supplied boolean/integer params before branching on them

erikrikarddaniel:
## PR checklist

## Description

Partial fix for #478: a boolean parameter cannot currently be turned **off** from the command line. `--skip_dbcan false` arrives as the *string* `'false'`, which is truthy in Groovy, so `if ( ! params.skip_dbcan )` stays false and the step silently stays skipped. `--min_contig_length` has the same root cause but a worse symptom: since it's compared with `> 0`, passing it on the CLI at all (e.g. `--min_contig_length 500`) crashes outright with `Cannot compare java.lang.String ... with java.lang.Integer`.

#478 originally proposed a full typed `params {}` block (83 params, requires Nextflow >=26.04, creates a second source of truth alongside `nextflow_schema.json`). We don't want to ship a release with the CLI bug still live, and don't want to wait on that larger migration or on a Nextflow version bump, so this is a much smaller, immediately-actionable partial fix.

I also looked at whether [nf-core/tools#4463](https://github.com/nf-core/tools/pull/4463) (bumps `nf-schema` and sets `cli_typecast`/`cast_cli_params` to enable automatic CLI type-casting) would fix this for free via a template sync -- it doesn't. I read nf-schema 2.7.3's actual source and confirmed live against a real pipeline run: `cast_cli_params`/`cli_typecast` only casts a **throwaway JSON copy** used for `validateParameters()`'s own validation step; the real `params` binding is never written back to, regardless of the setting. (I also confirmed `params.x = ...` reassignment inside a workflow body is a silent no-op, since `params` is a locked `ScriptBinding$ParamsMap` -- there's no cheap "coerce once and let it propagate" trick either.)

**What this PR does:** grepped the whole codebase for every param that's actually branched on with truthy/comparison semantics (as opposed to only ever interpolated into a shell command string, which is type-mismatch-safe either way). The entire surface turned out to be one file, `workflows/metatdenovo.nf`: 9 booleans (`annotate_only_consolidated`, `skip_dbcan`, `skip_eggnog`, `skip_eukulele`, `skip_fastqc`, `skip_kofamscan`, `skip_protein_consolidation`, `skip_qc`, `skip_trimming`) and 1 integer (`min_contig_length`). Everything else that looks bool-ish (`assembler`, `orf_caller`, `user_orfs_*`, `bbmap_ambiguous`) is schema-typed `string` and checked via non-empty-string truthiness, which isn't exposed to this bug. Every other numeric param is only ever interpolated into a command string. `cluster_min_seq_id` already defensively calls `.toString()` before `BigDecimal` parsing.

Added `typecastBooleanParam(name, value)`/`typecastIntegerParam(name, value)` next to the existing `validateInputSamplesheet` helper in `subworkflows/local/utils_nfcore_metatdenovo_pipeline/main.nf`. Each passes a real `Boolean`/`Integer` through unchanged, coerces the CLI strings `'true'`/`'false'` (or a numeric string), and **hard-errors** on anything else (e.g. `--skip_dbcan maybe`) instead of silently defaulting to truthy -- same UX the original issue's typed-params table proposed. Called once each at the top of `workflow METATDENOVO`'s `main:` in `workflows/metatdenovo.nf`, and every downstream read of those 10 params in that file uses the coerced local instead of `params.X`.

**Testing:** a standard pipeline `nf-test` can't exercise this, since it drives the pipeline via a params file (already correctly typed), not raw CLI strings. Instead added an `nf-test` `nextflow_function` suite (`subworkflows/local/utils_nfcore_metatdenovo_pipeline/tests/main.function.nf.test`) that calls the two functions directly -- 7 cases covering Boolean/Integer passthrough, string coercion both directions, and the hard-error path. Also verified end-to-end with a real invocation: `nextflow run . -profile test,docker -stub --skip_dbcan false` (the test profile sets `skip_dbcan = true`, so this CLI override exercises exactly the previously-broken path) -- confirmed `DBCAN:RUNDBCAN_DATABASE` now actually gets scheduled.

The rest of #478 (the full typed-`params{}` migration, or a real fix once/if nf-schema actually persists the CLI cast) is left for a future major version once Nextflow >=26.04 is an acceptable minimum.


nschan:
My understanding was that this should explicitly be avoided, i.e. all params should initialize to `false` and be truthed by `--paramname`, but I am unable to find documentation on that, so maybe it changed?

subworkflows/local/utils_nfcore_metatdenovo_pipeline/main.nf
```diff
@@ -224,6 +224,35 @@ def validateInputSamplesheet(input) {
...
+
+//
+// Nextflow's classic CLI parser hands boolean/integer params through as raw
+// strings (`--skip_dbcan false` -> the *string* 'false', which is truthy in
+// Groovy). nf-schema's validateParameters() does not fix this -- it only
+// type-casts a throwaway copy of params used for its own validation, never
+// the real `params` binding. Coerce explicitly at the read site instead.
+// See https://github.com/nf-core/metatdenovo/issues/478.
+//
+def typecastBooleanParam(String name, value) {
```
nschan: This could be turned into:
```suggestion
def typecastBooleanParam(String name) { 
    def value = params.get(name)
```

subworkflows/local/utils_nfcore_metatdenovo_pipeline/main.nf
```diff
@@ -224,6 +224,35 @@ def validateInputSamplesheet(input) {
...
+    if (value instanceof Boolean) {
+        return value
+    }
+    def s = value.toString()
+    if (s == 'true')  { return true }
+    if (s == 'false') { return false }
+    error("--${name} must be true or false, got '${value}'")
+}
+
+def typecastIntegerParam(String name, value) {
```
nschan: See above, same applies here

nschan (approved):
Generally fine, one suggestion to make it a bit smoother. There is also a comment ~below~ above, which questions the whole thing, but I am not sure if I am misremembering.

subworkflows/local/utils_nfcore_metatdenovo_pipeline/main.nf
erikrikarddaniel: Thanks, that simplifies things -- implemented!

subworkflows/local/utils_nfcore_metatdenovo_pipeline/main.nf
erikrikarddaniel: Dito

erikrikarddaniel:
> Generally fine, one suggestion to make it a bit smoother. There is also a comment ~below~ above, which questions the whole thing, but I am not sure if I am misremembering.

I'm also a bit in the dark here. This is the only way we (me and Claude) have gotten it to work so far. If it eventually turns out there's a better way, I'm happy to implement that.

PR #488, opened 2026-09-01
Adopt shared nf-core/modules custom/collectstats

erikrikarddaniel:
## PR checklist

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/metatdenovo/tree/master/docs/CONTRIBUTING.md)
- [ ] If necessary, also make a PR on the nf-core/metatdenovo _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [x] Make sure your code lints (`nf-core pipelines lint`).
- [x] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [ ] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [x] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).

## Description

Replaces the local `COLLECT_STATS` module with the shared `nf-core/modules` component
[`custom/collectstats`](https://github.com/nf-core/modules/tree/master/modules/nf-core/custom/collectstats),
the same swap nf-core/magmap already made in
[magmap#242](https://github.com/nf-core/magmap/pull/242). Addresses
[nf-core/magmap#237](https://github.com/nf-core/magmap/issues/237), which tracks sharing
summary-table generation code between the two pipelines.

No logic change in the R itself, but one visible output difference: the shared module
derives the feature-count column's name from its input filename rather than hardcoding
`n_feature_count`. For metatdenovo that means the column in
`<assembly>.<caller>.overall_stats.tsv.gz` is now named after the active ORF caller (e.g.
`prodigal`, `prokka`, `transdecoder`, `metaeuk`) instead of the generic `n_feature_count`.
All affected `nf-test` snapshots/assertions are updated to match, verified with real Docker
runs of every affected pipeline test.

Second commit is an unrelated one-line `.gitignore` fix (a local `.claude/` git-worktree
directory was showing up as untracked and blocking `nf-core pipelines lint --fix`).

_Reviewed and drafted with some help from Claude Code._

Issue #451, opened 2026-07-31
Record featureCounts' Unassigned_NoFeatures counts in overall_stats.tsv.gz


erikrikarddaniel:
### Description of feature

Add featureCounts' `Unassigned_NoFeatures` count (from the `*.featureCounts.tsv.summary` files that `SUBREAD_FEATURECOUNTS`/`FEATURECOUNTS_CDS` already produce) as a column in `overall_stats.tsv.gz` (built by `COLLECT_STATS`), alongside the existing `n_post_trimming`, `n_non_contaminated`, `idxs_n_mapped`/`idxs_n_unmapped`, and `n_feature_count` columns.

**Background**

This was originally investigated in nf-core/magmap (companion issue: nf-core/magmap#220), which shares an almost identical `COLLECT_STATS` implementation with this pipeline. The observation: a much larger fraction of reads map to contigs (`idxs_n_mapped`) than get assigned to any counted feature by featureCounts. Checking a real run's `*.featureCounts.tsv.summary` files there showed the entire "mapped but not counted" gap was explained by exactly two categories, `Unassigned_NoFeatures` and `Unassigned_Ambiguity`, with `NoFeatures` dominating by roughly an order of magnitude and every other category (`Unassigned_Chimera`, `Unassigned_MultiMapping`, `Unassigned_MappingQuality`, `Unassigned_Duplicate`, `Unassigned_Singleton`, `Unassigned_FragmentLength`) consistently at zero.

The same pattern (large `NoFeatures`, small everything-else) has been seen in a real nf-core/metatdenovo run too, which is arguably even more surprising here than in magmap: this pipeline assembles the transcriptome/genome directly from the sample's own reads rather than mapping against a pre-built genome/MAG collection, so there's less obvious reason to expect systematically incomplete annotation of the regions those same reads actually originate from. Not yet root-caused here specifically -- filing this primarily to get the diagnostic (`Unassigned_NoFeatures` as a visible column) in place so it's easier to investigate with real project data going forward, rather than only discoverable by opening individual per-sample `.summary` files.

**Implementation note**

When implementing this, it would be good to consolidate how this summary table is built across nf-core/magmap and nf-core/metatdenovo rather than maintaining near-duplicate local R scripts independently -- worth considering pulling the read-processing/stats-collection logic out into a shared nf-core subworkflow plus a set of modules that both pipelines (and others with the same pattern) can install and use.

erikrikarddaniel:
[...]

PR #504, opened 2026-09-10
Record featureCounts' Unassigned_* diagnostic counts in overall_stats.tsv.gz

erikrikarddaniel:
## PR checklist

## Description

Addresses #451: adds each featureCounts `Unassigned_*` category (`Unassigned_NoFeatures`,
`Unassigned_Ambiguity`, etc.) as its own column in `<assembly>.<caller>.overall_stats.tsv.gz`,
so the "mapped but not counted" gap described in the issue is visible without opening
individual per-sample `*.featureCounts.tsv.summary` files.

Note that #488/#489 (adopting the shared `custom/collectstats`/`custom/collectfeaturecounts`
nf-core/modules components) did **not** actually implement this despite being listed as
related work in #451's implementation note -- they were a mechanical port of the existing
local modules, and `FEATURECOUNTS_CDS.out.summary` was never wired into the workflow at all.
This PR is the actual feature.

**New local module**, `COLLECT_FEATURECOUNTSSUMMARY`: reshapes each sample's
`*.featureCounts.tsv.summary` (a wide table, one row per outcome category, one column per
BAM) into a narrow file per `Unassigned_*` category, run once per active ORF caller (same
grouping level as `CUSTOM_COLLECTFEATURECOUNTS`). `Assigned` is skipped, since it's already
the caller-named total column `CUSTOM_COLLECTFEATURECOUNTS`'s own output produces.

These files are fed into `CUSTOM_COLLECTSTATS` alongside each caller's existing counts
table -- it already derives a column name from the text between a file's first and second
dot, so `<caller>.<Status>.featureCounts.tsv` gives each category its own column for free.
One wrinkle found along the way: `CUSTOM_COLLECTSTATS` reads every `fcs` file it's handed in
one combined `read_tsv()` call, which requires them all to share the same columns -- so the
new files carry the same 9-column schema (`orf`/`chr`/`start`/`end`/`strand`/`length`/
`sample`/`count`/`tpm`) the counts table does, with only `sample`/`count` populated.

Verified against a real Docker run of `-profile test,docker` (not just the module-level
nf-test), confirming every `Unassigned_*` column actually lands correctly in
`megahit.prodigal.overall_stats.tsv.gz`.

Caught the same pre-existing stub-mode `TIDYVERSE_STRIPCDSPREFIX` bug already known from an
earlier PR (`gzip` in its `stub:` block fails if the target file is already staged in the
work dir) -- unrelated to this change, not fixed here.

_Reviewed and drafted with some help from Claude Code._


modules/local/collect/featurecountssummary/main.nf
```diff
@@ -0,0 +1,76 @@
...
+    # the same columns as CUSTOM_COLLECTFEATURECOUNTS's own counts table, even though only
+    # sample/count are populated here.
+    for ( s in unique(summaries\$status[str_starts(summaries\$status, 'Unassigned_')]) ) {
+        summaries %>%
+            filter(status == s) %>%
+            transmute(
+                orf = NA_character_, chr = NA_character_, start = NA_integer_, end = NA_integer_,
+                strand = NA_character_, length = NA_integer_, sample, count, tpm = NA_real_
+            ) %>%
+            write_tsv(paste0('${meta.caller}.', s, '.featureCounts.tsv'))
```
piplus2: **medium** — the output filename puts `meta.caller` before the first dot, but `CUSTOM_COLLECTSTATS` takes the column name from between the *first and second* dot (`'^[^.]+\\.([^.]+)\\..*'`). `meta.caller` is user-controlled and may contain dots: `schema_user_orfs.json` only enforces `^\S+$` on `name`, and `--user_orfs_name` has no `pattern` at all. With `--user_orfs_name myset.v2` every category file (`myset.v2.Unassigned_NoFeatures...`, `myset.v2.Unassigned_Ambiguity...`, ...) yields feature `v2`, so `group_by(sample, feature) %>% summarise(sum(count))` silently collapses all 13 categories into one column of their sum — no error, just wrong stats.

Fix: reject `.` in ORF-source names during the existing name validation in `workflows/metatdenovo.nf` (next to the collision/duplicate checks), or add a `pattern` to both schemas.

modules/local/collect/featurecountssummary/main.nf
```diff
@@ -0,0 +1,76 @@
...
+    library(stringr)
+
+    # CUSTOM_COLLECTSTATS names a table column from the text between a file's first and
+    # second dot, so name each output "<caller>.<Status>.featureCounts.tsv". Skip
+    # "Assigned": CUSTOM_COLLECTFEATURECOUNTS's own output already reports it, as the
+    # caller-named column.
+    summaries <- bind_rows(lapply(Sys.glob('*.featureCounts.tsv.summary'), function(f) {
+        read_tsv(f, col_types = cols(Status = col_character(), .default = col_integer())) %>%
+            rename(count = 2) %>%
+            mutate(sample = str_remove(basename(f), '\\\\.${meta.caller}\\\\.featureCounts\\\\.tsv\\\\.summary\$')) %>%
```
piplus2: **low** — `${meta.caller}` is interpolated raw into an R single-quoted literal here (and again at line 49), while `modules/nf-core/custom/collectstats/main.nf` already defines `rq()` for exactly this. A `--user_orfs` row named `it's` or `a\b` passes `^\S+$` validation and emits unparseable R. Route both interpolations through an `rq()`-style escape.

Also: `str_remove` returns its input unchanged on no-match, so if `task.ext.prefix` is ever set for `FEATURECOUNTS_CDS` this silently sets `sample` to the whole filename and every new column comes out all-`NA` instead of failing. Worth a `stopifnot()` that the pattern actually matched.

workflows/metatdenovo.nf
```diff
@@ -1054,17 +1069,29 @@ workflow METATDENOVO {
         // as a null meta and kill CUSTOM_COLLECTSTATS on tag "$meta.id" -- drop it instead, so a future
         // wiring mistake degrades to a missing stats row rather than a crash far from its cause.
         .filter { _caller, meta, _fcs, _mergetab -> meta != null }
-        .map { _caller, meta, fcs, mergetab -> [ meta, fcs, mergetab ?: [] ] }
+        // COLLECT_FEATURECOUNTSSUMMARY's output is `optional: true`, so a caller with no
+        // Unassigned_* rows never emits a tuple for it -- remainder: true, default to [].
+        .map { caller, meta, fcs, mergetab -> [ caller, meta, fcs, mergetab ] }
```
piplus2: **low (nit)** — this `.map` is an identity transform: it emits exactly its input. The comment above it describes the `remainder: true` behaviour of the `.join()` that follows, which makes the `.map` read as if it did something. Delete the line and keep the comment attached to the `.join`.

workflows/metatdenovo.nf
```diff
@@ -1054,17 +1069,29 @@ workflow METATDENOVO {
...
+        .join(
+            COLLECT_FEATURECOUNTSSUMMARY.out.unassigned.map { meta, unassigned -> [ meta.caller, unassigned ] },
```
piplus2: **low** — this join only ever matches callers that went through `CUSTOM_COLLECTFEATURECOUNTS`, so `protein_consolidate` (mixed into `ch_counts_per_caller` from `COLLECT_PROTEINCONSOLIDATE`) falls into the `remainder` branch and gets `[]`. Result: within one multi-caller run, `<assembly>.prodigal.overall_stats.tsv.gz` gains 13 columns while `<assembly>.protein_consolidate.overall_stats.tsv.gz` keeps the old 6, so the per-caller stats files are no longer column-compatible with each other.

Either feed `locus_consolidate`'s summaries through for the consolidated callers, or qualify the CHANGELOG entry, which currently promises the columns for `<assembly>.<caller>.overall_stats.tsv.gz` unconditionally.

tests/default.nf.test.snap
```diff
@@ -298,10 +304,10 @@
             4,
             349986,
             [
-                "sample\tn_post_trimming\tn_non_contaminated\tidxs_n_mapped\tidxs_n_unmapped\tprodigal"
+                "sample\tn_post_trimming\tn_non_contaminated\tidxs_n_mapped\tidxs_n_unmapped\tUnassigned_Ambiguity\tUnassigned_Chimera\tUnassigned_Duplicate\tUnassigned_FragmentLength\tUnassigned_MappingQuality\tUnassigned_MultiMapping\tUnassigned_NoFeatures\tUnassigned_NonSplit\tUnassigned_Overlapping_Length\tUnassigned_Read_Type\tUnassigned_Secondary\tUnassigned_Singleton\tUnassigned_Unmapped\tprodigal"
```
piplus2, line 307: **low** — this snapshot shows the change isn't purely additive: the caller column (`prodigal`, and `prokka`/`transdecoder` in the other snapshots) moves from position 6 to *last*, because `summarise()` sorts group keys and under `LC_COLLATE=C` uppercase `U` sorts before lowercase letters. Anyone reading `overall_stats.tsv.gz` positionally breaks.

Worth calling out in the CHANGELOG entry (which currently says only "alongside the existing ... columns") and in `docs/output.md` around line 335, which this PR leaves untouched.

piplus2:
just one comment (the medium one). the others are tiny stuff you can evaluate if worth it.

workflows/metatdenovo.nf
```diff
@@ -871,6 +880,56 @@ workflow METATDENOVO {
...
+        // in one invocation to share columns, so pad the reused diagnostic files to match
+        // before relabeling them. The original filenames are kept -- CUSTOM_COLLECTSTATS only
+        // reads the text between the first and second dot, so the "locus_consolidate" prefix
+        // is discarded either way.
+        ch_unassigned_protein_consolidate = ch_unassigned_per_caller.locus_consolidate
+            .flatMap { _meta, unassigned -> (unassigned instanceof List) ? unassigned : [ unassigned ] }
+            .collectFile { f ->
+                def lines = file(f).readLines()
+                def widened = ( [ lines[0] + '\tcallers\tn_calls\tn_loci\tloci' ] +
+                    lines[1..-1].collect { line -> line + '\t\t\t\t' } ).join('\n') + '\n'
```
piplus2: **medium** — this crashes on an empty file: `lines[1..-1]` on a zero-line list throws `java.lang.IndexOutOfBoundsException: fromIndex = -1` (confirmed by running this expression under Nextflow).

It's reachable in `-stub` runs: `COLLECT_FEATURECOUNTSSUMMARY`'s stub does `touch ${meta.caller}.Unassigned_NoFeatures.featureCounts.tsv`, `locus_consolidate` is created unconditionally (line 682), and the guard above passes under defaults (`skip_protein_consolidation = false`, `orf_callers` non-empty) — so `nextflow run ... -stub -profile test`, which `.github/workflows/download_pipeline.yml:106` runs, hits it. The real script never emits a header-only file, so this is stub-only.

Either short-circuit before indexing:

```groovy
if ( lines.size() < 2 ) return [ file(f).name, lines ? lines[0] + '\tcallers\tn_calls\tn_loci\tloci\n' : '' ]
```

or have the stub write a header row instead of `touch`.

erikrikarddaniel: Confirmed — reproduced the exact IndexOutOfBoundsException: fromIndex = -1 directly. Fixed by checking lines.size() < 2 (handles both the fully-empty stub case and the header-only/zero-rows case, which is also reachable in a real run) before the slice. Verified against a real -stub -profile test run: COLLECT_FEATURECOUNTSSUMMARY (megahit.locus_consolidate) now completes and the pipeline gets past this step cleanly — it only hits the separate, pre-existing TIDYVERSE_STRIPCDSPREFIX stub-mode gzip bug further along (unrelated to this PR, already noted elsewhere).

piplus2 (commented):
Reviewed all 6 commits. Wiring, `Unassigned_*` column padding against both the 9-col per-caller and 13-col `protein_consolidate` tables, `join(remainder: true)` handling and the `-stub`/header-only branch all check out. Two low-severity items:

### Low

**1. `meta.caller` is interpolated unescaped into an R regex** — [`modules/local/collect/featurecountssummary/main.nf#L41`](https://github.com/nf-core/metatdenovo/blob/issue-451-unassigned-nofeatures/modules/local/collect/featurecountssummary/main.nf#L41)
`--user_orfs_name` only rejects `.`, so a name like `orfs+v2` or `set[1]` fails to match and trips the `stopifnot`, and `orfs(v1` is a regex compile error — all after assembly/ORF calling/mapping have already run. The Groovy side already `Pattern.quote()`s caller names.
→ Match literally, e.g. `str_remove(basename(f), regex(paste0('.', caller, '.featureCounts.tsv.summary'), literal = TRUE))`, or `str_sub(f, 1, -nchar(suffix) - 1)` after a `str_ends(f, fixed(suffix))` check.

**2. CHANGELOG overstates the column position** — [`CHANGELOG.md#L10`](https://github.com/nf-core/metatdenovo/blob/issue-451-unassigned-nofeatures/CHANGELOG.md#L10)
"is now the last column instead" is only true with no annotation tables merged; with e.g. DIAMOND taxonomy the `diamondtax_*` columns still follow it (see `tests/diamond.nf.test.snap`). `docs/output.md` already says column order isn't fixed.
→ Reword to "…is no longer directly after `idxs_n_unmapped`; read columns by name."

https://claude.ai/code/session_01EDtXCi1jJqtvj5GgwCeS6b

erikrikarddaniel:
Both fixed and pushed.

**1. Regex injection**: confirmed -- and worse than I'd realized, `'orf(v1'` would be a regex compile error, not just a failed match, both only surfacing after assembly/ORF calling/mapping had already run. Fixed by matching the caller-derived suffix literally (`str_ends()` + `fixed()` to check, `str_sub()` to strip) instead of building a regex from it -- the caller name never reaches regex semantics at all now. Added a test with `orf+v2` as the caller name, confirmed it fails under the old logic and passes cleanly now.

**2. CHANGELOG wording**: reworded -- you're right that "last column" only holds with no annotation tables merged in. Points at column-name lookup instead now.

While fixing #1, swept the rest of the diff for the same class of issue (`meta.caller` reaching an unescaped interpolation context) rather than stopping at the two flagged spots. Found one more: the `stub:` block's `touch` command interpolated `meta.caller` raw into bash, so a name like `orf;rm -rf .` would break out of the command. Fixed the same way (single-quoted, standard bash quote-escaping), with a test using a semicolon in the caller name.

_Fixed with some help from Claude Code._

piplus2 (approved):
Re-reviewed `ca5168c` and `2697803` against the two low items from the previous pass — both addressed:

1. **Caller-name regex** — `modules/local/collect/featurecountssummary/main.nf` now strips the suffix with `str_ends(name, fixed(suffix))` + `str_sub()` instead of `str_remove(<regex>)`, so `orf+v2`/`set[1]`/`orfs(v1` no longer trip the `stopifnot` or fail to compile. Covered by the new "regex metacharacters is matched literally" nf-test. The stub's `touch` is also single-quoted (`safeCaller`), with a matching `-stub` test for `orf;v2`.
2. **CHANGELOG** — reworded to "no longer directly after `idxs_n_unmapped` -- read columns by name", which is accurate with annotation tables merged too.

CI is green on `2697803` (14/14 docker nf-test jobs, lint 316/316, pre-commit, nf-test-changes). The earlier six threads are all outdated by `34ba8ea`/`8a87a29`/`7b18b13`.

One non-blocking nit, fine to leave or fold into a follow-up: the comment just above the `stopifnot` (`main.nf` ~L48) still says "`str_remove()` returns its input unchanged on no match", but `str_remove()` is gone — the `else name` branch has the same no-match behaviour, so the guard is still correct, only the wording is stale.

Approving.

https://claude.ai/code/session_01CQSvKBC66C9nYKQjX7uQUX

erikrikarddaniel:
Thanks for your patience @piplus2 !

Issue #476, opened 2026-08-26
Add a stub-mode pipeline test to cover channel wiring without databases


erikrikarddaniel:
## Why

The EggNOG bug fixed on the [#460](https://github.com/nf-core/metatdenovo/issues/460) branch annotated only the *first* ORF caller: with `--orf_caller prokka,transdecoder` plus protein consolidation, three protein sets reached eggnog-mapper and two of them were silently skipped.
Nothing in the test suite could catch it.
`test_eggnog`, `test_kofamscan` and `test_dbcan` are all single-caller with `skip_protein_consolidation = true`, so one annotation task is the correct answer there, and the profiles that *are* multi-caller skip every annotation tool because the databases are too large for CI.
It took a full-size run to surface, and the only symptom was two missing `*.emapper.tsv.gz` files and a missing `n_eggnog` column.

A `-stub` run executes each process's `stub:` block instead of the real tool -- no databases, no real compute, seconds per task -- while exercising the real channel topology.
A multi-caller stub run with the annotation tools enabled would have failed on this bug immediately, because eggnog-mapper would have run once where three tasks were expected.

## Stub mode nearly works already

Measured, not assumed.
`-profile test,docker -stub` runs the whole default pipeline to completion (74 tasks) once two things are fixed, and `--orf_caller prodigal,transdecoder` with protein consolidation on completes 92 tasks.

What stands in the way:

1. **The MultiQC ORF-statistics blocks crashed on an empty protein FASTA** with `ERROR ~ f != java.lang.Integer`, because the no-ORFs branch handed an integer `0` to `String.format('%.1f', ...)`.
   Already fixed on the #460 branch; listed here only because it is the first thing a stub run hits.

2. **`modules/nf-core/seqkit/grep`'s stub is broken upstream.**
   It builds `suffix = output_type + compression_suffix` and then appends another `.gz`, so gzipped input produces `*.fa.gz.gz`, which does not match the module's declared `path("*.{fa,fq,fa.gz,fq.gz}")` and fails as a missing output.
   This is a real bug in nf-core/modules and deserves a PR there, with a local `nf-core modules patch` in the meantime.

3. **`modules/local/eggnog/download`'s stub is incomplete.**
   It does not create `eggnog_proteins.dmnd`, and its `ln -s eggnog/* ./` collides with the module's `storeDir`, so the task exits 1 with `mv: cannot overwrite ...: Directory not empty`.

[...]

PR #506, opened 2026-09-11
Add a -stub pipeline test covering multiple ORF callers with every annotation tool on

erikrikarddaniel:
## PR checklist

## Description

Closes #476.

A `-stub` run exercises the real channel topology without databases or real compute (seconds per task), but nothing in the existing test suite ran one with more than one ORF caller and any annotation tool active at once -- exactly the combination where a channel-wiring bug can silently drop a protein set from an annotation tool instead of erroring, since `-stub` produces output regardless of what actually reached it. That's exactly what happened on the `#460` branch: with `--orf_caller prokka,transdecoder` plus protein consolidation, three protein sets should have reached eggnog-mapper and two were silently skipped, only caught by a full-size real run.

`conf/test_stub.config` + `tests/stub.nf.test` runs the default pipeline with `--orf_caller prokka,transdecoder`, protein consolidation on (`--annotate_only_consolidated false`, so every caller's own protein set reaches annotation, not just the consolidated one), and EGGNOG/KOFAMSCAN/dbCAN/EUKulele/HMMER all enabled. It asserts each of those gets exactly one task per active protein set (prokka, transdecoder, the mmseqs-clustered consolidation), rather than snapshotting file contents -- `-stub` mode makes file contents meaningless, so the task-count assertion is where this test's value actually is.

Fixes five real `-stub` bugs found while building it (measured, not assumed -- iterated against a real `-stub` run until it completed cleanly), none reachable from any other existing test profile:

- `eukulele/format_tax` and `prokka/gff2tsv` had no `stub:` block at all, so `-stub` ran their real R scripts against the real upstream script's own stubbed-out (empty) input and crashed.
- `kofamscan/unique`'s stub wrote `${prefix}.unique.tsv.gz`, which doesn't match its own declared output glob `*-uniq.tsv.gz` at all.
- `eggnog/download`'s stub never created `eggnog_proteins.dmnd`, one of its four declared outputs, and its `ln -s`/`storeDir` combination could collide on a rerun.
- `tidyverse/stripcdsprefix`'s stub output filename could collide with its own staged input's filename (a symlink from `storeDir`), and `gzip` refuses to overwrite an existing file.

Also pulls in [nf-core/modules#12798](https://github.com/nf-core/modules/pull/12798)'s upstream fix for `seqkit/grep`'s stub appending `.gz` twice to gzipped input (`nf-core modules update seqkit/grep`).

No new CI job needed: `.github/workflows/nf-test.yml` already auto-discovers and shards every `tests/*.nf.test` file, so `tests/stub.nf.test` is picked up by the existing matrix for free.

---


danilodileo (approved):
LGTM, it was time to update those stub!

Issue #510, opened 2026-09-13
Flaky snapshot: software-versions YAML inner-map key order is non-deterministic


erikrikarddaniel:
### Description of the bug

Every pipeline-level `tests/*.nf.test` snapshots `pipeline_info/nf_core_metatdenovo_software_mqc_versions.yml` (via `removeNextflowVersion(...)`) as part of its assertion. The key order of the *inner* per-process tool-version maps (e.g. `COLLECT_FEATURECOUNTSSUMMARY: { R, readr, dplyr, stringr }`) is not stable across otherwise-identical runs, causing intermittent `Different Snapshot` failures with no actual content difference -- every key/value pair is present and correct, just reordered.

Found while investigating CI failures on PR #509, all 7/7 `docker | 25.10.4` shards failing while every `latest-everything` shard passed. Verified this is a genuine pre-existing flake, not something specific to that PR:

- The *raw*, per-task `versions.yml` (written directly by each R script at task execution time, e.g. `COLLECT_FEATURECOUNTSSUMMARY`'s own `writeLines(c(...))` call) is always correct and deterministic -- confirmed by inspecting the actual work-dir file from a failing run, which had the *correct* order (`R, readr, dplyr, stringr`) even though the final combined YAML did not.
- `processVersionsFromYAML()` (`subworkflows/nf-core/utils_nfcore_pipeline/main.nf`, vendored nf-core template code) round-trips each process's yaml through SnakeYAML (`yaml.load(...)` -> `collectEntries` -> `yaml.dumpAsMap(...)`). Tested this function in isolation (10 repeated calls on identical input, via a standalone Nextflow script) and it is perfectly deterministic on its own.
- So the reordering happens somewhere in the workflow-level aggregation between the per-task file and the final `collectFile(sort: true, ...)`-written combined file -- most likely a concurrency-related effect of Nextflow's dataflow engine processing multiple version entries through that shared transform concurrently, though the exact line hasn't been pinned down with full certainty.
- Confirmed this reproduces on a completely clean, unmodified `upstream/dev` checkout too (not just on a feature branch), ruling out it being specific to any one PR's code changes -- though a given PR's own runtime-timing changes elsewhere in the pipeline can be enough to flip which non-deterministic ordering a given run happens to land on, which is presumably why this hasn't been noticed as a *recurring* problem before now.

### Command used and terminal output

```console
$ NXF_VER=25.10.4 nf-test test tests/megahit_prokka.nf.test --profile=+docker

[...]

PR #509, opened 2026-09-13
Fix --user_orfs_gff silently losing MetaEuk-only loci

erikrikarddaniel:
## PR checklist

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/metatdenovo/tree/master/docs/CONTRIBUTING.md)
- [ ] If necessary, also make a PR on the nf-core/metatdenovo _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [x] Make sure your code lints (`nf-core pipelines lint`).
- [x] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [ ] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [x] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).

## Description

Closes #508.

`--user_orfs_gff`/`--user_orfs_faa` feeds a user-supplied gff straight into `FORMAT_GFF2BED`, skipping the `FORMAT_METAEUK_GFF`/`FORMAT_METAEUKFAA` normalisation a pipeline-internal `--orf_caller metaeuk` run always gets first. Found while diagnosing a real production run: MetaEuk's own resume/cache broke on an 82M-locus assembly, so its already-completed raw output was re-supplied via `--user_orfs_gff`/`--user_orfs_faa` instead of re-running MetaEuk. `FORMAT_GFF2BED`'s `(^|;)ID=` anchor never matches raw `TCS_ID=`/`Target_ID=` attributes, so every one of that caller's ~42M records silently got an empty-string id -- and because locus consolidation's member-dedup keys on that literal (identical) string, every one of those distinct genomic loci collapsed into a single fake locus in the output tables. The only visible symptom was `FORMAT_LOCUSFAA` failing much later with "1 of 82192853 loci had no matching protein sequence" -- the "1" being the fake merged locus, not the real number of affected genes.

Two independent fixes:

[...]

PR #511, opened 2026-09-13
Fix flaky software-versions snapshot ordering in pipeline-level tests

erikrikarddaniel:
## PR checklist

## Description

Closes #510.

Every pipeline-level `tests/*.nf.test` snapshots `pipeline_info/nf_core_metatdenovo_software_mqc_versions.yml`. The key order of a given process's *inner* tool-version map (e.g. `COLLECT_FEATURECOUNTSSUMMARY: { R, readr, dplyr, stringr }`) is not stable across otherwise-identical runs, causing intermittent `Different Snapshot` failures with no actual content difference -- every key/value pair present and correct, just reordered.

Found while investigating CI failures on PR #509 (all 7/7 `docker | 25.10.4` shards failing while every `latest-everything` shard passed). Verified this is a real, pre-existing flake independent of that PR's own changes:

- The *raw*, per-task `versions.yml` (written directly by each module's own script at task execution time) is always correct and deterministic -- confirmed against the actual work-dir file from a failing run.
- `processVersionsFromYAML()` (vendored nf-core template code) is deterministic in isolation -- tested with 10 repeated calls on identical input via a standalone Nextflow script.
- Reproduces on a completely clean, unmodified `dev` checkout too, not just on a feature branch.

So the reordering happens somewhere in the workflow-level aggregation between the per-task file and the final combined YAML (most likely a concurrency-related effect of Nextflow's dataflow engine processing multiple version entries through that shared transform concurrently), not in any individual module's own script.

Rather than patch vendored aggregation code without being certain of the exact mechanism, this PR sorts the parsed structure recursively before snapshotting in every pipeline-level test -- guaranteed effective regardless of root cause, and doesn't touch what end users actually see in the real output file.

Every regenerated snapshot's content was verified structurally identical to its predecessor (deep JSON comparison, order-insensitive) before committing -- only key order changed, nothing else.

Also bumps `nft-utils` 0.0.3 -> 1.2.0 in `nf-test.config` (brings `sanitizeOutput` and other fixes available for future test work; no existing local test currently uses functions from this plugin, so no snapshot content is affected by the bump itself).

This PR was developed with the help of Claude Code, including diagnosing the root cause and verifying every regenerated snapshot's content equivalence.


tests/spades_prodigal.nf.test
```diff
@@ -18,13 +18,26 @@ nextflow_pipeline {
             def stable_name = getAllFilesFromDir(params.outdir, relative: true, includeDir: true, ignore: ['pipeline_info/*.{html,json,txt}', 'multiqc/multiqc_plots', 'multiqc/multiqc_plots/**' ])
             // stable_path: All files in ${params.outdir}/ with stable content
             def stable_path = getAllFilesFromDir(params.outdir, ignoreFile: 'tests/.nftignore')
+            // The order of keys within a given process's inner tool-version map is not
```
jonasscheid: afiak nft-utils@0.0.3 → 1.2.0 is enough. You dont need this funtion, can you check?

jonasscheid (changes requested):

tests/spades_prodigal.nf.test
erikrikarddaniel: Thanks! Perfect if it is, I'll check.

erikrikarddaniel: Confirmed — verified against the nft-utils source directly: `removeFromYamlMap()` (which `removeNextflowVersion()` calls) wraps every map level in a `TreeMap` as of the current 1.2.0, so the bump alone produces sorted keys. Dropped the `sortMapDeep` wrapper, regenerated snapshots, and confirmed the full 47-test suite still passes. Thanks for the catch!

(Reviewed/fixed with some help from Claude Code.)

jonasscheid (approved):
👍🏼

PR #515, opened 2026-09-14
Patch release 1.4.1

### LINKED ISSUE #513: Is `bgzip` missing in BBNorm's Singularity container?
### Description of the bug

I got this Slack report:

Hi, I am running an nf-core/metatdenovo workflow on the HPC system, and the BBNorm step fails after approximately 41 hours while trying to write its normalized FASTQ output.
The job successfully completed the k-mer counting/hash-table construction:

```
BBNorm memory allocation: 819 GB
Hash table: ~598 GB
Estimated unique k-mers: ~211 billion
Hash-table occupancy: 86.08%
Table creation time: ~147,067 s (~41 h)
The process then fails with:
java.io.IOException: Cannot run program “bgzip”:
/scratch/SlurmTMP/userID.23778641/TEMPFILE_BBNORM_P1_R1_11cf4e18dc999d13e904e68b23d0cc__62875.239315649414.fq.gz
(No such file or directory)

Caused by: java.io.FileNotFoundException:
/scratch/SlurmTMP/userID.23778641/TEMPFILE_BBNORM_P1_R1_11cf4e18dc999d13e904e68b23d0cc__62875.239315649414.fq.gz
(No such file or directory)

Exception in thread “main” java.lang.AssertionError:
Could not execute bgzip -c -f -l 4 -@ 6
The BBNorm command is:
bbnorm.sh
 in=<~100 FASTQ files>
 out=all_samples.fastq.gz
 target=50
 min=5
 threads=6
 -Xmx819g
```

The process is running inside the nf-core/metatdenovo Singularity container. I have verified that bgzip is present inside the container:
/opt/conda/bin/bgzip
bgzip (htslib) 1.21
and /opt/conda/bin is in the container PATH.
BBNorm itself does not explicitly set a temporary directory. The error indicates that it is attempting to use:
/scratch/SlurmTMP/userID.23778641/
for a temporary FASTQ file.
An initial check shows no node or filesystem problems on the node used.
Could this be a Singularity problem (i.e. that you have to mount/connect the external directories to some internal container overlays)?

### Command used and terminal output

```console

```

### Relevant files

_No response_

### System information

_No response_

erikrikarddaniel:
## Description

This targets `master` rather than `dev` intentionally: it's a candidate for a `1.4.1` patch
release, so users hitting this bug on `1.4.0` can pick up the fix without pulling in `dev`'s much
larger, cache-invalidating changeset.

Fixes #513: BBNorm fails under Singularity/Apptainer with a misleading
`Cannot run program "bgzip"` / `FileNotFoundException` error on a path like
`/scratch/SlurmTMP/<user>.<jobid>/TEMPFILE_BBNORM_P1_R1_...fq.gz`, even though `bgzip` is present
and working inside the container.

**Root cause:** Nextflow's Singularity/Apptainer launcher forwards the host's `$TMPDIR` into the
container unconditionally (`SINGULARITYENV_TMPDIR`), but only bind-mounts the task's own work
directory. On HPC systems where Slurm sets `$TMPDIR` to a per-job scratch directory, that path
doesn't exist inside the container. BBNorm writes its multipass intermediate files to `$TMPDIR`,
so the write fails with `ENOENT`, which Java's `ProcessBuilder` then surfaces as the misleading
`bgzip` error. Confirmed by inspecting the reporter's own `.command.run`, which shows exactly this
forwarding and bind-mount pattern.

**Fix:** pass `tmpdir=.` to `bbnorm.sh`, BBNorm's own documented option to redirect multipass temp
files to the task's own (always bind-mounted) work directory instead of relying on `$TMPDIR`. (An
initial attempt at `usetempdir=f`, also advertised in BBTools' generic help text, turned out not
to be a recognized parameter for this tool — caught by running the real test suite rather than
trusting the help text.)

**On test coverage:** a true failing-then-passing regression test isn't practical for this module.
Nextflow's Docker launcher never forwards `TMPDIR` at all, so the bug is invisible under the
module's normal (Docker) CI. Under Singularity, with an intentionally unmounted `TMPDIR`, the
module's existing small test fixtures still don't reach BBNorm's spill-to-disk path even when
forced into a multipass run (`passes=2`) — the mechanism only triggers at the data/runtime scale
of the original report (~100 input files, ~41h runtime). Verified manually against a Singularity
profile instead.

Also bumps the pipeline version to `1.4.1` (`nf-core pipelines bump-version 1.4.1`) and pre-patches
the pipeline-level `nf-test` snapshots' embedded version string accordingly, since this PR is meant
to become the `1.4.1` release directly.

**Unrelated CI fix included here:** this PR's own `nf-test` checks were repeatedly failing
`conda | 25.10.4 | 2/5` with `InvalidArchiveError(...[Errno 28] No space left on device)` while
building the `MULTIQC` conda environment — the self-hosted RunsOn `nf-test` runner exhausts its
default 30GB disk. `dev` already carries the fix (bump the runner's `volume` label to `40gb`,
matching [nf-core's documented example](https://nf-co.re/docs/developing/testing/ci-runners/#increasing-disk-space-for-ci-runners)),
but it was never brought over to `master`. Included directly here (rather than as a separate PR)
so this PR's own CI can actually go green without waiting on a second full review cycle.

## PR checklist

- [ ] If you've fixed a bug or added code that should be tested, add tests! — see "On test coverage" above for why this isn't practical here.
- [x] Ensure the test suite passes (`nextflow run . -profile test_bbnorm,docker --outdir <OUTDIR>`).


Joon-Klaps (approved):

erikrikarddaniel:
Superseded by #517 — same commits, but opened from a branch pushed directly to `nf-core/metatdenovo` instead of the fork, so the AWS full-size test can actually get its secrets/variables (fork-PR `pull_request_review` runs don't get them, which is why this PR's AWS check kept failing empty in ~14s no matter what).

Issue #474, opened 2026-08-26
Sorted BAM files and their indexes are published regardless of --save_bam


erikrikarddaniel:
## What happens

`--save_bam` defaults to `false`, but every run still publishes a sorted BAM and its index to `results/samtools/`.
On the full test that is ~700 MB per sample across four samples, i.e. several GB of alignments and indexes in the published output of a run whose results are linked from the documentation.

## Why

There are two BAM publish paths, and only one of them is gated.

`BBMAP_ALIGN` ([`conf/modules.config#L129-L144`](https://github.com/nf-core/metatdenovo/blob/0c421761f5d5f8d33b99f69509fb340c80a37cf0/conf/modules.config#L129-L144)) publishes its unsorted `*.bam` with `enabled: params.save_bam`, so with the default nothing lands in `bbmap/bbmap/`.

The `.*SAMTOOLS.*` block ([`conf/modules.config#L150-L171`](https://github.com/nf-core/metatdenovo/blob/0c421761f5d5f8d33b99f69509fb340c80a37cf0/conf/modules.config#L150-L171)) then publishes `*.bam` and `*.bam.bai` with **no `enabled:` guard at all**:

```groovy
withName: ".*SAMTOOLS.*" {
    publishDir = [
        [
            pattern: "*.bam",
            path: { "${params.outdir}/samtools/" },
            mode: params.publish_dir_mode

        ],
        [
            pattern: "*.bam.bai",
            path: { "${params.outdir}/samtools/" },
            mode: params.publish_dir_mode
        ],
        [
            pattern: "*.{flagstat,idxstats}",
            ...
            enabled: params.save_samtools,
        ]
    ]
}
```

Only the `{flagstat,idxstats}` rule in that block is gated, on `params.save_samtools` (default `true`).
So suppressing the unsorted BBMap BAM just means a sorted copy of the same alignments gets published instead, which is not what the param's own description ("Save the bam files from mapping") says.

This is long-standing rather than a recent regression — `git log -L 150,171:conf/modules.config` puts the block at ecd5f42 ("Merge 3.2.1 template plus some more").

## Documentation gap

[`docs/output.md#L242`](https://github.com/nf-core/metatdenovo/blob/0c421761f5d5f8d33b99f69509fb340c80a37cf0/docs/output.md#L242) documents `bbmap/bbmap/*.bam` as being written only "if enabled with `--save_bam`", which is the misleading half of the story.
[...]

PR #519, opened 2026-09-15
Gate SAMTOOLS BAM/BAI publishDir on --save_bam

danilodileo:
## Summary

Fixes #474.

The `.*SAMTOOLS.*` `publishDir` block in `conf/modules.config` published `*.bam`/`*.bam.bai` unconditionally, with no `enabled:` guard. As a result, setting `--save_bam false` (the default) only suppressed `BBMAP_ALIGN`'s own unsorted BAM — a sorted copy of the same alignments, plus its index, still got published to `<outdir>/samtools/` on every run regardless of the flag, several GB per full-test run.

- Added `enabled: params.save_bam` to the `*.bam` and `*.bam.bai` publishDir rules in the `.*SAMTOOLS.*` block, matching `BBMAP_ALIGN`'s existing `*.bam` rule and the param's own description ("Save the bam files from mapping").
- Left the `*.{flagstat,idxstats}` rule as-is, gated on `--save_samtools` (on by default), unaffected by this change.
- Documented the `samtools/` output directory in `docs/output.md` (it previously had no entry).
- Regenerated all pipeline-level nf-test snapshots affected by the removed BAM/BAI entries.

## Out of scope

`SAMTOOLS_STATS`'s `*.stats` output has a separate, pre-existing publishDir gap (not part of this fix) — left as-is per the issue's own scoping.

## Test plan

- [x] `prek run -a` passes
- [x] Regenerated and manually diffed all 13 pipeline-level nf-test snapshots (`default`, `annotate_only_consolidated`, `bbmap_ambiguous`, `megahit_prokka`, `megahit_prokka_prodigal`, `megahit_prokka_transdecoder`, `megahit_transdecoder`, `metaeuk`, `metaeuk_transdecoder`, `single`, `spades_prodigal`, `diamond`, `kofamscan`) — each diff shows only the expected removal of `samtools/*.sorted.bam`/`.sorted.bam.bai` entries, with `flagstat`/`idxstats` entries retained


erikrikarddaniel (approved):
👍

Issue #520, opened 2026-09-16
BBMap-family modules should all use a memory headroom multiplier on -Xmx, like BBNorm does


erikrikarddaniel:
## What happens

The AWS full-scale test run failed in `BBMAP_ALIGN (ERR978615)` with exit status 1 and no useful error surfaced to Nextflow/Tower (the real `bbmap.sh` stderr goes to `${prefix}.bbmap.log` via `&>`, which Tower doesn't show). The task died ~5 minutes after starting, consistent with a memory-related failure rather than a genuine alignment error.

## Why this looks like the BBNorm problem again

`BBMAP_ALIGN`, `BBMAP_BBDUK` and `BBMAP_INDEX` all set:

```groovy
-Xmx${task.memory.toGiga()}g
```

i.e. the exact task memory ceiling, with zero headroom for JVM overhead (metaspace, native buffers, BBMap's own off-heap hashing structures) below the container's hard memory limit.

`BBMAP_BBNORM` used to do the same thing, and we hit this exact failure mode with it before. We fixed it with a headroom multiplier on the `-Xmx` value (`modules/nf-core/bbmap/bbnorm/main.nf`):

```groovy
memory = "-Xmx${Math.round(Math.max(1, Math.floor(task.memory.toGiga() * 0.8)))}g"
```

Git history on that module shows this was tuned empirically, not guessed once: `ec5d6bf Reduce memory to bbnorm to 0.95` first, then `b7d3e28 Give bbnorm only 0.8` after 0.95 still wasn't enough headroom. The multiplier pattern itself actually originates upstream in nf-core/modules' own `bbmap/bbnorm` (default `0.95`); our local patch (`modules/nf-core/bbmap/bbnorm/bbmap-bbnorm.diff`) only tightens it to `0.8`.

`BBMAP_ALIGN` (`modules/nf-core/bbmap/align/main.nf`), `BBMAP_BBDUK` (`modules/nf-core/bbmap/bbduk/main.nf`, already locally patched for an unrelated reason) and `BBMAP_INDEX` (`modules/nf-core/bbmap/index/main.nf`) never got the same treatment and still use the unguarded `task.memory.toGiga()`.

## Suggested fix

[...]

PR #521, opened 2026-09-16
Give BBMAP_ALIGN/BBDUK/INDEX the same memory headroom as BBNORM

danilodileo:
## Summary

Fixes #520.

`BBMAP_ALIGN`, `BBMAP_BBDUK` and `BBMAP_INDEX` all set `-Xmx${task.memory.toGiga()}g` — the exact task memory ceiling, with zero headroom for JVM overhead (metaspace, native buffers, BBMap's own off-heap hashing structures) below the container's hard memory limit. `BBMAP_BBNORM` hit this exact failure mode before and was patched to use a fraction of `task.memory` instead; the other three tools never got the same treatment.

- Applied the same headroom-multiplier pattern to all three, via a local `nf-core modules patch` on each module (matching how `BBMAP_BBNORM`'s own patch already works).
- Used **0.9** as the multiplier for all three (vs. BBNorm's 0.8): none of these tools has actually been observed failing this way before, unlike BBNorm, so less headroom is taken away as a starting point.
- Kept as a local patch rather than an upstream nf-core/modules PR — the right value is something we expect to want to retune per-tool based on our own experience, and getting BBNorm's own value changed upstream previously turned into a lengthy discussion.

## Verification

- [x] `prek run -a` passes
- [x] Module-level nf-test: `BBMAP_BBDUK` and `BBMAP_INDEX` pass unchanged
- [x] `BBMAP_ALIGN`'s module tests fail in this environment, but reproduced identically against the **unmodified** module — a pre-existing gap (a `bam()` nf-test helper unavailable here), unrelated to this change
- [x] Manually inspected the rendered `.command.sh` for both `BBMAP_ALIGN` and `BBMAP_INDEX` test runs to confirm the `-Xmx` substitution produces a sane, scaled-down value relative to `task.memory`

## Note

The AWS full-test failure that prompted #520 is still not 100% confirmed as a memory issue — nobody has pulled Seqera Platform's per-task resource graph or the actual `ERR978615.bbmap.log` yet. This PR closes the underlying inconsistency (one BBMap tool patched, three not) regardless of whether it's the exact cause of that particular failure.


erikrikarddaniel (approved):
AI-assisted review (Claude, on behalf of @erikrikarddaniel).

Code matches the agreed approach in #520: same guarded-fallback `-Xmx` headroom pattern `BBMAP_BBNORM` already carries (`-Xmx3g` default if `task.memory` is unknown, else `Math.floor(task.memory.toGiga() * 0.9)`), applied consistently as local `nf-core modules patch` diffs across `bbmap/align`, `bbmap/bbduk`, `bbmap/index`. `modules.json` and CHANGELOG correctly updated.

The one red CI shard (`docker | 25.10.4 | 1/7`, on `tests/stub.nf.test`) is unrelated to this PR — it's the software-versions snapshot key-order mismatch you diagnosed in #522 (stale local nf-test 0.9.3 snapshots vs CI's pinned 0.9.4), not caused by anything here.

Approving.

PR #524, opened 2026-09-17
Update bbmap/bbnorm to fix Singularity/Apptainer TMPDIR bug

erikrikarddaniel:
## PR checklist

- [x] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/metatdenovo/tree/master/docs/CONTRIBUTING.md)
- [ ] If necessary, also make a PR on the nf-core/metatdenovo _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [x] Make sure your code lints (`nf-core pipelines lint`).
- [ ] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [ ] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [x] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).

## Description

Closes #516.

Updates the vendored `bbmap/bbnorm` module to pick up [nf-core/modules#12934](https://github.com/nf-core/modules/pull/12934), which fixes BBNorm failing under Singularity/Apptainer with a misleading `bgzip`/`FileNotFoundException` error when the host's `$TMPDIR` isn't bind-mounted into the container. `bbnorm.sh` now writes its multipass temp files to `tmpdir=tmp` under the task's own (always bind-mounted) work directory instead.

This is the same bug already hotfixed directly on `master` for the 1.4.1 release (#517, closing #513) via a hand-written local patch. `dev` deliberately didn't carry that local patch forward -- this PR brings `dev` the real upstream fix instead, as planned when 1.4.1 shipped.

This pipeline's own `-Xmx` headroom patch (0.8 multiplier instead of upstream's 0.95, see #520/#521) no longer applied cleanly on top of the upstream change, so it was reapplied by hand against the new module base and the `.diff` regenerated with `nf-core modules patch`.

[...]

PR #525, opened 2026-09-17
Bundle awscli with EGGNOG/KOFAMSCAN/dbCAN download containers

danilodileo:
## Summary

Fixes #471.

`EGGNOG_DOWNLOAD`, `KOFAMSCAN_DOWNLOAD` and `RUNDBCAN_DATABASE` each cache their downloaded database in a `storeDir`, which `-profile test_full` points at an `s3://` path. On AWS Batch, Nextflow stages that `storeDir` copy by shelling out to `aws` from inside the task's own container -- none of these three had it, so the download itself succeeded and the run then died with `aws: command not found` copying it into the store.

- Built combined Wave containers (each tool + `awscli`, both docker and singularity variants) via the `wave-cli` tool, and verified locally that both binaries actually run inside each one before using them.
- Pointed each module's `container` directive at the new image and added `awscli` to its `environment.yml`. `RUNDBCAN_DATABASE` is a vendored nf-core module, so its change is tracked via `nf-core modules patch`.
- Re-enabled dbCAN in `conf/test_full.config` now that the underlying cause is fixed -- it was only skipped as a workaround for this bug. EggNOG was already enabled (and so has presumably been failing in `awsfulltest` runs the same way).
- Left KOfamScan skipped for now: it works the same way, but isn't currently pointed at an `s3://` `storeDir`, so it wasn't hitting this bug -- enabling it is out of scope here.

## Second, unrelated bug in the same area

`workflows/metatdenovo.nf`'s `eukulele_dbpath` existence check used `java.io.File`, which can't resolve a remote path like `s3://` and was silently creating a bogus local `s3:` directory in the launch dir instead of doing anything useful. Switched to Nextflow's own `file()`, matching how `channel.fromPath()` a few lines below already handles the same param correctly.

## Verification

- [x] `prek run -a` passes
- [x] Verified `aws` and each module's own tool (`download_eggnog_data.py`/`wget`, `wget`, `run_dbcan`) both run correctly standalone inside each new container
- [ ] The actual AWS Batch `storeDir`-to-S3 staging behavior can't be verified locally -- local Docker/Singularity execution doesn't exercise this code path at all (Nextflow's own head process handles S3 staging natively there; only AWS Batch tasks, running as isolated remote jobs, stage via the in-container `aws` CLI). Real verification needs `awsfulltest` running against real AWS infrastructure, which triggers automatically once this PR has two reviewer approvals.


modules/local/eggnog/download/main.nf
```diff
@@ -3,9 +3,13 @@ process EGGNOG_DOWNLOAD {
     label 'process_low'
 
     conda "${moduleDir}/environment.yml"
+    // Bundles awscli alongside eggnog-mapper: this process' storeDir can point at an s3:// path
+    // (e.g. --eggnog_dbpath in conf/test_full.config), and Nextflow's AWS Batch executor stages
+    // storeDir output to S3 by shelling out to `aws` from inside the task's own container --
+    // which the plain eggnog-mapper image doesn't have, addresses #471.
```
erikrikarddaniel:
```suggestion
    // storeDir can point at an s3:// path; Nextflow's AWS Batch executor stages that copy by
    // shelling out to `aws` inside the task's own container, which the plain eggnog-mapper image lacks.
```

modules/local/kofamscan/download/main.nf
```diff
@@ -3,9 +3,13 @@ process KOFAMSCAN_DOWNLOAD {
     label 'process_long'
 
     conda "${moduleDir}/environment.yml"
+    // Bundles awscli alongside wget: this process' storeDir can point at an s3:// path (via
+    // --kofam_dir), and Nextflow's AWS Batch executor stages storeDir output to S3 by shelling
+    // out to `aws` from inside the task's own container -- which the plain wget image doesn't
+    // have, addresses #471.
```
erikrikarddaniel:
```suggestion
    // storeDir can point at an s3:// path; Nextflow's AWS Batch executor stages that copy by
    // shelling out to `aws` inside the task's own container, which the plain wget image lacks.
```

modules/nf-core/rundbcan/database/main.nf
```diff
@@ -2,9 +2,13 @@ process RUNDBCAN_DATABASE {
     label 'process_low'
 
     conda "${moduleDir}/environment.yml"
+    // Bundles awscli alongside dbcan: this process' storeDir can point at an s3:// path (via
+    // --dbcan_dbpath), and Nextflow's AWS Batch executor stages storeDir output to S3 by
+    // shelling out to `aws` from inside the task's own container -- which the plain dbcan image
+    // doesn't have, addresses #471.
```
erikrikarddaniel:
```suggestion
    // storeDir can point at an s3:// path; Nextflow's AWS Batch executor stages that copy by
    // shelling out to `aws` inside the task's own container, which the plain dbcan image lacks.
```

workflows/metatdenovo.nf
```diff
@@ -996,8 +996,11 @@ workflow METATDENOVO {
     // SUBWORKFLOW: Eukulele
     //
     if ( ! skip_eukulele ) {
-        // Make sure the eukulele_dbpath exists
-        d = new File("${params.eukulele_dbpath}")
+        // Make sure the eukulele_dbpath exists. Resolved with Nextflow's own file(), not
+        // java.io.File, since eukulele_dbpath can be a remote path (e.g. s3://) that only
+        // Nextflow's NIO-backed Path understands -- a plain File silently treats it as a
+        // nonexistent local path and creates a bogus local directory tree instead.
```
erikrikarddaniel:
```suggestion
        // Make sure the eukulele_dbpath exists. file() (not java.io.File) resolves remote paths like s3://.
```

erikrikarddaniel (approved):
Nice fix -- this matches the root cause exactly (the aws call only fires when AWS Batch stages the storeDir copy, which is why it never showed up locally). Verifying the tool + awscli standalone in each new image is the right substitute for a real AWS Batch run.

Two small things, neither blocking:

1. This PR conflicts with dev, but only in CHANGELOG.md. #524 merged this morning and reordered the Fixed section (it had drifted: #521 above #523 despite #523 being the higher/newer PR number). This branch predates that reorder, so GitHub's CONFLICTING status is just that ordering diff, not a real conflict in the fix itself -- a rebase/merge onto current dev should resolve it trivially.
2. The three module comments (eggnog/kofamscan/rundbcan main.nf) each end "...addresses #471". Per our standing "trim comments" convention, the storeDir/AWS-Batch-staging fact is worth keeping (it's a real, non-obvious gotcha), but the issue-number framing rots the moment #471 isn't in front of the reader, and each could be a line or two shorter. Suggestions below for all three, plus the eukulele_dbpath one which doesn't reference an issue but can also be tightened to one line.

Issue #473, opened 2026-08-25
Also write summary tables as Parquet


erikrikarddaniel:
## Summary

Also write the `summary_tables/` outputs as Parquet, alongside the gzipped TSVs.

The tables are the pipeline's actual deliverable and are meant to be read straight into R or
Python. On a real project they get large -- the counts tables are one row per feature per sample --
and TSV is a poor format for that: no types, no column pruning, no predicate pushdown, and slow to
read repeatedly. Parquet fixes all four and costs little to produce.

## Existing implementation to port

We already wrote an nf-core module for exactly this, `duckdb/table2parquet`, and it is in use in
nf-core/magmap. Nothing new needs writing; this is a port.

The module takes what the summary tables already are:

```groovy
input:
tuple val(meta), path(table)

output:
tuple val(meta), path("*.parquet"), emit: parquet
```

magmap's wiring is directly reusable (`workflows/magmap.nf`):

```groovy
if ( save_parquet ) {
    DUCKDB_TABLE2PARQUET(
        <table channels>
            .mix(COLLECT_FEATURECOUNTS.out.counts.map { _meta, tsv -> tsv })
            .mix(COLLECT_STATS.out.overall_stats)
            .map { tsv -> [ [ id: tsv.name.replaceAll(/\.tsv(\.gz)?$/, '') ], tsv ] }
    )
}
```

plus a `withName: DUCKDB_TABLE2PARQUET` block publishing `*.parquet` to `summary_tables/`, and a
boolean param:

```json
"save_parquet": {
  "type": "boolean",
  "description": "Also write the summary tables in Parquet format, alongside the default gzipped TSV.",
  "fa_icon": "fas fa-table"
}
```

magmap passes `save_parquet` down through the `take:` block rather than reading `params` inside the
workflow, which matches this pipeline's convention too.

## What is different here

metatdenovo has considerably more to convert than magmap -- 12 `withName` blocks publish to
`summary_tables/`. Worth deciding whether to convert all of them or only the ones that actually get
large:

- **Clearly worth it** (one row per feature per sample): `<assembly>.<orfcaller>.counts.tsv.gz`,
  `<assembly>.locus_consolidate.counts.tsv.gz`,
  `<assembly>.protein_consolidate_<identity>.counts.tsv.gz`.
- **Probably worth it** (one row per feature, sometimes several): the annotation tables --
[...]

PR #526, opened 2026-09-17
Also write summary tables as Parquet (--save_parquet)

erikrikarddaniel:
## PR checklist

- [ ] If necessary, also make a PR on the nf-core/metatdenovo _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository. (not needed -- no new fixture data)
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`). (not run)
- [ ] Usage Documentation in `docs/usage.md` is updated. (not applicable -- a single self-explanatory boolean flag, covered in `docs/output.md` instead)
- [x] `README.md` is updated (including new tool citations and authors/contributors).

## Description

Closes #473.

Adds `--save_parquet` (default `false`): when set, every `summary_tables/` tsv is also written as Parquet, alongside the unaffected default gzipped TSV. Ported from nf-core/magmap's existing `DUCKDB_TABLE2PARQUET` wiring, using the shared `nf-core/modules` component `duckdb/table2parquet`.

**Scope decision:** the issue raised converting all 12 `summary_tables/`-publishing `withName` blocks vs. only the ones that get large. Went with converting everything -- with `--save_parquet` on, every table is uniformly Parquet-backed, so a `duckplyr`-style lazy read doesn't get forced to materialize early just because one table in a join happens to still be TSV-only. Off by default, so this costs nothing for existing users.

**Wiring:** most of these `summary_tables/`-producing processes (annotation subworkflows) only run behind their own `skip_*` flag, so there's no single unconditional channel to mix from. An accumulator channel (`ch_parquet_tables`) collects each optional subworkflow's own table from inside its own `if` block, and gets combined with the always-invoked tables (counts, locus/protein consolidation, HMM classification, Diamond taxonomy, overall stats) in one final `DUCKDB_TABLE2PARQUET` call, re-keyed on each table's own filename rather than its upstream `meta` (several of these channels' metas collide, e.g. multiple callers sharing an id, while the published filenames are already unique) -- same idiom magmap already uses.

**Verification beyond the checklist above:**

- `tests/stub.nf.test` (`test_stub` profile, multi-caller + every annotation tool) now also runs with `--save_parquet` on, and asserts exactly 26 `DUCKDB_TABLE2PARQUET` invocations -- one per `summary_tables/` tsv that profile's active tools produce, worked out by hand and confirmed against the actual trace.
- `tests/default.nf.test` (`-profile test`, real execution) also runs with `--save_parquet` on, exercising real `duckdb` SQL against real gzipped TSVs, not just `-stub` touches.
- Both pass; version/file-listing snapshots regenerated to include the new `duckdb` version entry and `.parquet` outputs.


danilodileo (approved):
👍

Issue #365, opened 2025-04-08
Allow Nextflow to natively handle kofamscan download


edmundmiller:
https://github.com/nf-core/metatdenovo/blob/master/modules/local/kofamscan/download/main.nf

The way this is done might break for systems where the compute nodes don't have access to download the files in this process.

Using `file()` and then unzip/untar process would be a better way of handling this.

erikrikarddaniel:
Verified this works, not just in theory.

Built a minimal Nextflow process with `storeDir` set, running inside a container with `--network none` (fully network-isolated, to actually simulate a compute node that can't reach the internet), and a normal declared `path` input built from `file(url)` at the workflow level -- no `wget`/`curl` in the script at all, just `tar`/`gunzip`.

Result: Nextflow logs `Staging foreign file: <url>` and fetches it as a distinct step before the task is even submitted -- confirmed the file was genuinely staged and usable inside the task despite it having zero network access itself. A second run with `storeDir` already populated skipped the whole process entirely (fetch included), same caching behaviour as today.

So the fix here is: drop the `wget` calls from `KOFAMSCAN_DOWNLOAD`'s script, build the two inputs (`ko_list`, `profiles`) via `file(ko_list_url)`/`file(profiles_url)` at the workflow level instead of passing the URLs in as `val`, declare them as normal `path` inputs, and keep the existing `gunzip`/`tar` extraction plus `storeDir` as-is.

While checking this, found the same fixable pattern in `modules/local/eggnog/download/main.nf` -- three separate `wget` + `gunzip`/`tar` calls, and that one doesn't even have `storeDir` set today, so it's arguably in worse shape (re-downloads every fresh work dir regardless of network access). Worth bundling into the same fix.

`modules/local/metaeuk/download/main.nf` downloads via `metaeuk databases` (the tool's own built-in fetcher) rather than raw `wget`, so it has the same underlying network-access concern but isn't fixable with this technique -- that's the tool's internal logic, not something Nextflow's staging can intercept. Separate problem if it ever needs solving.

_(Looked into with some help from Claude Code.)_

erikrikarddaniel:
Fixed in #527

PR #527, opened 2026-09-18
Use Nextflow's native file() staging instead of wget in KOFAMSCAN_DOWNLOAD/EGGNOG_DOWNLOAD

erikrikarddaniel:
## PR checklist

## Description

Addresses #365: `KOFAMSCAN_DOWNLOAD` and `EGGNOG_DOWNLOAD` shelled out to `wget`
from inside the task itself, which breaks on compute nodes without network
access (e.g. HPC compute nodes).

Fix: build the download URLs into `path` inputs with Nextflow's own `file()` at
the call site, instead of passing the URL strings into the process as `val`.
Nextflow's own staging then fetches the URL before the task is submitted, so
the task itself never needs network access. Verified this actually works
end-to-end in a `--network none` container (see the issue comment).

While making this change, hit a second, unrelated bug it exposed: Nextflow
always stages `path` inputs as symlinks, and GNU gzip's in-place `gunzip`
opens with `O_NOFOLLOW`, which fails with `ELOOP` ("Too many levels of
symbolic links") on a symlinked target even though there's no real loop. So
`gunzip <staged-file>` in place no longer works once the input is a real
Nextflow `path` rather than a plain locally-`wget`ed file — fixed by streaming
to stdout instead (`gunzip -c <staged-file> > outfile`), which follows the
symlink fine. Root-caused with `strace` inside the actual task container
(`open(..., O_NOFOLLOW) = -1 ELOOP`), not just worked around.

Applied the same fix to `EGGNOG_DOWNLOAD`, which had the identical `wget`+
in-place-`gunzip` pattern (previously noted in the issue as worth bundling in).

**Testing:**
- `tests/kofamscan.nf.test` (`-profile test_kofamscan,docker`, real execution,
  not stub) passes end-to-end with the new staging — this exercises
  `KOFAMSCAN_DOWNLOAD` for real against the small test-datasets KOfam
  subset, confirming both the `file()` staging and the `gunzip -c` fix work
  together on a real run, not just in isolation.
- `EGGNOG_DOWNLOAD` has no real (non-stub) pipeline test in this repo today,
  so its `gunzip -c`/symlink fix was verified by reproducing the exact
  ELOOP failure and its fix directly inside the module's own container
  (`community.wave.seqera.io/library/eggnog-mapper_awscli`), not just by
  reasoning about it.
- `nf-core pipelines lint`, `prek run -a`, `nextflow lint` (both the declared
  minimum Nextflow version and the latest installed) all clean; pre-existing
  unrelated warnings (module version bumps, the known
  `utils_nextflow_pipeline` `JsonGenerator.Options` lint error) untouched by
  this change.


erikrikarddaniel:
Pushed a follow-up fix while doing the full-scale \`test_eggnog\` verification this PR's own testing section calls for: \`EGGNOG_DOWNLOAD\`'s hardcoded host was \`eggnog6.embl.de\`, which redirects to a dead path. The correct host — confirmed against eggnog-mapper's own upstream \`download_eggnog_data.py\` and verified all three files actually serve real content there — is \`eggnog5.embl.de\`. This is a pre-existing typo, not something introduced by the \`file()\`-staging change; it was never caught before because \`EGGNOG_DOWNLOAD\` has no real (non-stub) pipeline test.

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Changed`
 
+- [#NN](https://github.com/nf-core/metatdenovo/pull/NN) - `KOFAMSCAN_DOWNLOAD` and `EGGNOG_DOWNLOAD` now stage their reference-database URLs via Nextflow's own `file()` instead of shelling out to `wget`, so the download works even when the compute node running the task has no network access itself; also fixes `EGGNOG_DOWNLOAD`'s download host, which had silently been pointing at a dead URL (`eggnog6.embl.de` instead of `eggnog5.embl.de`), addresses [#365](https://github.com/nf-core/metatdenovo/issues/365) (@erikrikarddaniel)
```
danilodileo: Change log needs to be updated

danilodileo (approved):
👍

CHANGELOG.md
erikrikarddaniel:
```suggestion
- [#527](https://github.com/nf-core/metatdenovo/pull/527) - `KOFAMSCAN_DOWNLOAD` and `EGGNOG_DOWNLOAD` now stage their reference-database URLs via Nextflow's own `file()` instead of shelling out to `wget`, so the download works even when the compute node running the task has no network access itself; also fixes `EGGNOG_DOWNLOAD`'s download host, which had silently been pointing at a dead URL (`eggnog6.embl.de` instead of `eggnog5.embl.de`), addresses [#365](https://github.com/nf-core/metatdenovo/issues/365) (@erikrikarddaniel)
```

PR #530, opened 2026-09-19
Fix five bugs found reviewing the 2.0 release diff

erikrikarddaniel:
## PR checklist

## Description

Five bugs found reviewing the full `master...dev` diff ahead of the 2.0 release.
All five are in code added during this cycle, so none has shipped and there is no CHANGELOG entry.

### 1. A normal Prokka GFF is rejected as raw MetaEuk output

Whether a GFF is raw MetaEuk output is decided from its feature lines.
`hasIdAttribute()` read the first 20 lines and counted the header toward that limit.

A GFF3 header is one `##gff-version` line plus one `##sequence-region` **per contig**, so past ~19 contigs no feature line falls in the window.
The GFF is then classified as raw while the matching `.faa` is not, and the run aborts with

> USER_ORFS: <id>'s gff and faa disagree on whether they're raw MetaEuk output or already normalised

So `--user_assembly` with `--user_orfs_gff`/`--user_orfs_faa` from Prokka fails immediately on any realistic assembly, blaming the user's inputs for a mismatch that does not exist.

Fixed by skipping comment lines before the limit applies, which is what the neighbouring `.faa` check already did.
A regression test is added; it fails with that exact error when the fix is reverted.
The three existing tests all use a GFF with a zero-line header, so none of them could have caught this.

### 2. `TIDYVERSE_STRIPCDSPREFIX` overwrites its upstream task's cached output

A process must not write to its own staged input.
`write_tsv("${prefix}.counts.tsv.gz")` names the exact file `CUSTOM_COLLECTFEATURECOUNTS` emitted — a symlink into that task's work directory — and follows it.

So the upstream task's cached output is rewritten on every successful run and its stored hash stops matching its content.
Reproduced in the module's own container: upstream md5 `39d75fa9…` before, `cd8433ef…` after.
Same class as nf-core/modules#12799.

Now writes `stripped.counts.tsv.gz` and `file.rename()`s it into place, replacing the symlink instead of writing through it — which is what the module's own `stub:` block already did.

### 3. Protein consolidation silently skipped for `--user_orfs_gff`/`--user_orfs_faa`

Five guards tested `( orf_callers || params.user_orfs )`.
Neither is set for `--user_assembly` plus `--user_orfs_gff`/`--user_orfs_faa`, a combination the validation explicitly accepts, so protein consolidation is skipped although `--skip_protein_consolidation` was not given.
The same expression also gated the `--bbmap_ambiguous all` validation and three downstream channels.

All five now use `user_orf_names`, which is built from `--user_orfs` rows *and* `--user_orfs_name`.

### 4. `protein_consolidate_<pct>` was not a reserved caller name

`reserved_caller_names` held only `locus_consolidate`, while its own comment said "names" plural.
`protein_consolidate_name` was computed 50 lines below the validation, so a `--user_orfs` row could claim it, giving two entries with the same `meta.caller` — and the `.join()` keyed on `meta.caller` then drops one.
The computation moves above the validation that needs it.

### 5. Six new local modules used the pre-Apptainer container guard

`workflow.containerEngine == 'singularity'` is false under Apptainer, so those modules fall through to the Docker URI.
The other new modules in this release already use `in ['singularity', 'apptainer']`.

The 16 **pre-existing** local modules carrying the same guard are fixed here too, so all 31 `modules/local` modules now match. Closes #531.

Safe to include rather than defer: CI already runs `-profile singularity`, so every one of these Singularity URIs is pulled and run on each pass.
The change only routes Apptainer to that same proven image instead of the Docker one.

### Verification

`nf-test` on `subworkflows/local/user_orfs` — 5/5 pass, 1 red with the fix reverted.
`nf-core pipelines lint` 726 passed / 0 failed, `prek run -a` clean, `nextflow lint` clean apart from the pre-existing vendored `utils_nextflow_pipeline` `JsonGenerator.Options` error.


modules/local/tidyverse/stripcdsprefix/main.nf
```diff
@@ -26,10 +26,14 @@ process TIDYVERSE_STRIPCDSPREFIX {
     library(dplyr)
     library(stringr)
 
+    # The output name matches the staged input, which is a symlink. Write elsewhere and rename,
+    # or write_tsv follows the symlink and overwrites the upstream task's cached file.
     read_tsv("${counts}", show_col_types = FALSE) %>%
         # Transdecoder appends "cds." to ORF IDs in the gff file, but does not in the fasta file. Remove to make compatible between tables.
         mutate(orf = str_remove(orf, '^cds\\\\.')) %>%
-        write_tsv("${prefix}.counts.tsv.gz")
+        write_tsv("stripped.counts.tsv.gz")
```
danilodileo, line 34: isn't better some sort of variables instead of a fixed string? it probably doesn't matter here but it would be good to keep consistency across modules

danilodileo (approved):
💯

modules/local/tidyverse/stripcdsprefix/main.nf
erikrikarddaniel: It's a scratch name that only exists inside the task; the next line renames it to `${prefix}.counts.tsv.gz`, which is what gets published. Writing straight to the final name would follow the staged input symlink and overwrite the upstream task's cached file.

PR #535, opened 2026-09-20
Keep a spliced gene in a single locus during locus consolidation

erikrikarddaniel:
## PR checklist

### The bug

Every ORF should belong to exactly one locus. A spliced gene could end up in two.

`FORMAT_GFF2BED` writes one BED interval per CDS record, so a spliced gene arrives as one interval per exon.
`FORMAT_LOCUSCONSOLIDATE` sweeps those intervals in coordinate order, groups the ones that overlap, and names each group as it closes.
When another caller has calls in the intron, a gene's two exons do not close in the same group:

- exon 1 does not overlap the next interval, so its group closes alone and is named after that ORF;
- exon 2 overlaps the group the other callers have built, and that group is named from its coordinates.

Two names for one gene.
The members table maps an ORF to one locus, so only the second name keeps it.
The first locus is left with no protein, which is what `FORMAT_LOCUSFAA` reports on a full-scale run:

```
ERROR: 126974 of 89927258 loci had no matching protein sequence
```

Counts are wrong too, and silently: the gene's reads are divided between the two loci.
Only a splice-aware caller can trigger this, since Prokka and TransDecoder write one CDS record per gene.

### The fix

`flush()` keeps a map from ORF to locus for the current contig.
When a group closes, its members are looked up there and the group joins the locus it finds instead of naming a new one.
A locus keeps the name its first group gave it, so a locus with several contributors can be named after one of its ORFs rather than from coordinates; the provenance table is what says how many callers agreed.

Groups are also closed when the contig changes rather than at `END`, and a closed group's state is freed.
That cuts peak memory about fourfold on real input, and gives the last group on each contig a defined emission order -- `END` iterated an awk array, whose order is undefined.

No CHANGELOG entry: locus consolidation (#467, #479) has never shipped in a release.

### Verification

- The module's awk over 5M BED lines from a real run: 21492 ORFs in more than one locus before, 0 after.
- Two new module tests, one for each order the exons can merge in. Both fail without the fix; the four existing tests pass either way.
- `tests/metaeuk_transdecoder.nf.test`, `tests/megahit_prokka_prodigal.nf.test` and `tests/megahit_prokka_transdecoder.nf.test` pass with no snapshot change.
- `prek run -a`, `nf-core pipelines lint` and `nextflow lint` clean.


danilodileo:
why not reporting it in changelog?

danilodileo (approved):
👍

erikrikarddaniel:
> why not reporting it in changelog?

Because this function was never part of a release, so the change is not in relation to anything.

